PR #13, opened 2020-07-24
Swap table implementation

alan-mj-lin:
- Swap Request table added to database

implemented APIs:
- GET swap-request
- POST swap-request
- PATCH swap-request/accept
- PATCH swap-request/approve
- DELETE swap-request


showduhtung (approved):
I think there could be more conditions, but I can't think of it, and I think when it comes up, we can just add to it. Otherwise it looks mostly fine to me.

PR #14, opened 2020-08-03
swap notifications

alan-mj-lin:
- task table updates
- tasks now sort by date
- swap notifications table and APIs
- timeZone feature added


server/routes/swap-notifications.ts
```diff
@@ -0,0 +1,158 @@
...
+        }
+    },
+);
+
+router.post(
+    '/swap-notifications',
+    async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            jwt.verify(req.headers.authorization, cert);
+            let message = '';
```
showduhtung: can you move this message variable to right above the switch case in line 83

server/routes/swap-notifications.ts
```diff
@@ -0,0 +1,158 @@
...
+                case 'approved':
+                    message = `Your request to switch with ${requesteeUser.firstName} has been approved`;
+                    break;
+                case 'cancelled':
+                    message = `Your request to switch has been cancelled`;
+                    break;
+                case 'created':
+                    message = `Request sent to ${requesteeUser.firstName}`;
+                    if (swapRequest.type === 'requestOne') {
+                        await db.SwapNotification.create({
```
showduhtung, line 98: I think I have a general understanding of what's going on. Can you make a commented out message/essay that describes what this does?

showduhtung:
Looks pretty good to me, but do you think you could test all of these functions with ted and the front end first before we merge this pull request?

shared/SequelizeTypings/models/ChurchModel.ts
```diff
@@ -5,6 +5,7 @@ export interface ChurchAttributes {
     name: string;
     address: string;
     description: string;
+    timeZone: string;
```
showduhtung: lets make this timezone

PR #17, opened 2020-09-06
all server changes for notifications to work properly

alan-mj-lin:


server/db/models/Token.ts
```diff
@@ -1,7 +1,10 @@
 import * as Sequelize from 'sequelize';
 import { SequelizeAttributes } from 'shared/SequelizeTypings/typings/SequelizeAttributes';
 import { TokenInstance, TokenAttributes } from 'shared/SequelizeTypings/models';
-import helper from '../../helper_functions';
+
+function addMinutes(date: Date, minutes) {
```
showduhtung, line 5: this could be put in a shared utilities folder

server/helper_functions.ts
```diff
@@ -106,13 +107,42 @@ const funcs = {
...
     setDate(date: string, time: string, timeZone: string) {
         return DateTime.fromISO(`${date}T${time}`, { zone: timeZone });
     },
+    async sendPushNotification(
+        userId: number,
+        userPushToken: string,
+        title: string,
+        body: string,
+    ) {
+        const user = await db.SwapNotification.findAll({
```
showduhtung, line 124: It looks like what you're finding is all swap notifications that has to do with the userId. So the result will be a notification[]? Instead of naming it user, maybe something like unreadNotifications makes more sense?

server/helper_functions.ts
```diff
@@ -106,13 +107,42 @@ const funcs = {
...
+            where: { id: userId, isRead: false },
+            attributes: ['isRead'],
+        });
+        const badgeNumber = user.length + 1;
+        fetch('https://exp.host/--/api/v2/push/send', {
```
showduhtung, line 129: do you mind helping me understand what this does?

server/routes/swap-notifications.ts
```diff
@@ -54,32 +55,34 @@ router.post(
     async (req: Request, res: Response, next: NextFunction) => {
         try {
             jwt.verify(req.headers.authorization, cert);
+            const decodedToken = jwt.decode(req.headers.authorization, { json: true });
```
showduhtung, line 58: so looking at this `router.post()` code, it looks like you have some `const` variables hoisted before your switch case logic, but what I noticed is that some of your constants aren't being used after the switch case, which means there are times when it's not being used. You'll see if you comment out the switch case body, your code editor should tell you that `const localChurchUsers` and `const requesteeUser` are not being used after the switch case, which is pretty expensive seeing that you're doing a `db.find()` possibly for no reason.

I recommend moving `const localChurchUsers` and `const requesteeUser` into the switch case when needed

server/routes/swap-notifications.ts
```diff
@@ -116,10 +132,16 @@ router.post(
                         .send({ message: 'Invalid notification type.' });
             }
             await db.SwapNotification.create({
-                userId: req.body.userId,
+                userId: requestingUser.id,
```
showduhtung, line 135: I think you've explained this to me before but one more time? If it goes through the switch case and gets `'created'`, it creates a `SwapNotification` and does a `helper.sendPushNotification`, and does it again after the switch case?

server/routes/swap-notifications.ts
```diff
@@ -132,14 +154,77 @@ router.post(
     },
 );
 
+router.patch(
+    '/swap-notifications/:notificationId',
+    async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            jwt.verify(req.headers.authorization, cert);
```
showduhtung, line 161: what happens if the jwt is unverified? does it keep going?

server/routes/swap-requests.ts
```diff
@@ -89,18 +92,47 @@ router.post('/swap-requests', async (req: Request, res: Response, next: NextFunc
...
+            const newRequest = await db.SwapRequest.findOne({
+                where: { id: createRequest.id },
+                include: [
+                    {
+                        model: db.Task,
+                        as: 'task',
+                        attributes: ['id', 'userId'],
+                    },
+                ],
+            }).then(async (request) => {
```
showduhtung, line 125: @whoahs  making an axios request from our own server to another part of our server, what do you make of this?

server/routes/tasks.ts
```diff
@@ -16,21 +16,19 @@ module.exports = router;
 
 router.get('/tasks', async (req: Request, res: Response, next: NextFunction) => {
     try {
-        // jwt.verify(req.headers.authorization, cert);
+        jwt.verify(req.headers.authorization, cert);
         const searchArray = [];
```
showduhtung, line 20: maybe we can call this searchParamList?

server/routes/users.ts
```diff
@@ -18,50 +20,49 @@ module.exports = router;
...
-                include: [
-                    {
-                        model: db.Church,
-                        as: 'church',
-                        attributes: ['name'],
-                    },
-                ],
+        const decodedToken = jwt.decode(req.headers.authorization, { json: true });
+        const loggedInId = decodedToken.sub.split('|')[1];
+        const searchArray = [];
```
showduhtung, line 25: searchParamList

server/routes/users.ts
```diff
@@ -18,50 +20,49 @@ module.exports = router;
...
             });
-            if (users) res.status(200).json(users);
-            else res.status(404).send({ message: 'Not found' });
+            searchArray.push({ id: userIds });
         }
+        const searchParams = {
+            [Op.and]: searchArray,
+            [Op.not]: { id: loggedInId },
+        };
+        console.log(searchParams);
```
showduhtung, line 45: comment this out

PR #19, opened 2020-10-09
St/update seed

showduhtung:


showduhtung:
- A lot of the file changes is from abstracting out logic and making lots of new files from them
- Improved the script file
- Added more data to seed file
- got started on new tables (probably should've done it in a separate pull request)

server/routes/notifications.ts
```diff
@@ -0,0 +1,188 @@
...
+            attributes: ['id', 'requesteeUserId', 'type', 'accepted', 'approved'],
+            include: [
+                {
+                    model: db.Task,
+                    as: 'task',
+                    attributes: ['id', 'userId'],
+                },
+            ],
+        });
+        const { requesteeUserId, task, type: swapRequestType } = request;
```
hchiam, line 78: Where is `swapRequestType` defined? it's used 4 times in this file

hchiam:
Reviewed 75% of the files. It'll take me a while to review the rest of the bigger diffs. I see you're reworking some table relationships, but I'm not up-to-date on the DB setup, so hopefully @CedThen and/or @alan-mj-lin can double-check that stuff.

hchiam (changes requested):

server/routes/notifications.ts
showduhtung: so this is a es6 thing where you can destructure and then rename it. 
```
const message = { greeting: 'Hello' };
const { greeting: renamedVariable }  = message;

console.log(renamedVariable + 'World')
// 'Hello World'
```

showduhtung: so it's being defined in line 77

hchiam: oh cool! i thought it was a type ahha

scripts/makeDatabase.js
```diff
@@ -0,0 +1,110 @@
+/* eslint-disable array-callback-return */
+/* eslint-disable no-unused-vars */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable no-plusplus */
```
hchiam: fighting the linter i see :(
Suggestions:
1. add `"no-plusplus": "off"` in the .eslintrc (so i++ etc works inside basic for loops). Not sure why it's not covered by all the `recommended`s in `extends`.
2. This file looks like it's just for testing, but if you want to clean it up, you can replace the 2 offending `.map`s with `.forEach`, so you can remove `eslint-disable array-callback-return`
```suggestion
/* eslint-disable no-unused-vars */
/* eslint-disable @typescript-eslint/no-unused-vars */
```

scripts/makeDatabase.js
```diff
@@ -0,0 +1,110 @@
...
+const fs = require('fs');
+const { users: databaseUsers, roles: databaseRoles, events: databaseEvents } = require('./webDatabase');
+
+const userToId = () => {
+    const mappedUsers = {};
+    databaseUsers.map((user, index) => {
```
hchiam: This is what I mean. Here: 
```suggestion
    databaseUsers.forEach((user, index) => {
```

scripts/makeDatabase.js
```diff
@@ -0,0 +1,110 @@
...
+        mappedUsers[user.name] = index + 1;
+    });
+    return mappedUsers;
+};
+
+const roleToId = () => {
+    const mappedRoles = {};
+    databaseRoles.map((role, index) => {
```
hchiam: and here: 
```suggestion
    databaseRoles.forEach((role, index) => {
```

server/db/models/User.ts
```diff
@@ -88,9 +85,7 @@ const UserFactory = (
     User.beforeUpdate(createSaltyPassword);
 
     User.associate = (models) => {
-        User.hasMany(models.Task, { foreignKey: 'UserId' });
-        User.belongsTo(models.Church, { as: 'church', foreignKey: 'ChurchId' });
-        User.belongsToMany(models.Role, { through: models.UserRole, as: 'role' });
+        User.belongsTo(models.Church, { as: 'church', foreignKey: 'churchId' });
     };
```
hchiam: Interesting. So Instead of directly associating user with role and task, you're associating user with user_role with task. Kinda like a linear `user -> user_role -> task` instead of parallel arrows `user -> (role, task)`.

server/routes/authentication.ts
```diff
@@ -1,387 +1,274 @@
...
+        const { email: userEmail, password, isVerified, id } = user;
+        jwt.verify(authorization, password);
+
+        const [message, status] =
+            // eslint-disable-next-line no-nested-ternary
+            userEmail !== queryEmail
+                ? ['Invalid Request', 401]
+                : isVerified
+                ? ['Password change success.', 201]
+                : ['User does not exist or is not verified', 401];
```
hchiam: interesting. very concise and avoids using `let`

server/routes/requests.ts
```diff
@@ -0,0 +1,192 @@
...
+        const { userId: yourUserId } = await db.UserRole.findOne({ where: { id: yourTask.userRoleId } });
+
+        const myTask = await db.Task.findOne({ where: { id: myTaskId } });
+        const { userId: myUserId } = await db.UserRole.findOne({ where: { id: myTask.userRoleId } });
+
+        if (!myTask) res.status(404).send({ message: 'Task not found' });
+        await myTask.update({ status: 'changeRequested' });
+
+        const request = await db.Request.create({
+            requesteeUserId: yourUserId,
```
hchiam: i personally found the `my`s and `your`s a little ambiguous until I read this line. We could make it more obvious/boring with 
```suggestion
            requesteeUserId,
```
and then replace `your...` -> `requestee...` and `my...` -> `requester...` or `requesting...` (unless I have the POV backwards)

server/routes/requests.ts
```diff
@@ -0,0 +1,192 @@
...
+        return res.status(503).send({ message: 'Server error, try again later' });
+    }
+});
+router.delete('/requests/:requestId', certify, async (req: Request, res: Response, next: NextFunction) => {
+    try {
+        const request = await db.Request.findOne({
+            where: { id: req.params.requestId },
+            attributes: ['id', 'requesteeUserId', 'type', 'accepted', 'approved', 'taskId'],
+        });
+        return request ? res.status(200) : res.status(404).send({ message: 'Swap request not found' });
```
hchiam: the `DELETE` route isn't actually performing a delete for now, right?

server/routes/requests.ts
```diff
@@ -0,0 +1,192 @@
...
+        const { accepted, approved, task, id: requestId } = request;
+        const { userId } = await db.UserRole.findOne({ where: { id: task.userRoleId } });
+        const [message, status] = determineMessageStatus(request, accepted, approved);
+
+        // does this need to test for requestAll?
+        if (status === 202 && !approved && accepted) {
+            await request.update({ id: requestId, approved: true });
+            postNotification(requestId, userId, 'approved', message, req.headers.authorization);
+        }
+        return status === 202 ? res.status(status).json(request) : res.status(status).send({ message });
```
hchiam: Maybe yes? If I understand correctly, the requestee currently can hit to "accept" _all_ requested task swaps from one requester _at once_, but currently the requestee _cannot_ "approve" all tasks from a requester all at once? I'm not super clear what the distinction between accept/approve is in this app. Is "accept" more like viewing, or are they both buttons but "approve" is more like hitting an "are you sure" confirmation button?

server/routes/tasks.ts
```diff
@@ -1,225 +1,163 @@
...
+                where: { id: task.userRoleId },
+                attributes: ['id'],
             });
-            res.status(200).send({ message: 'Task replacement successful.' });
-            // }
+
+            const [message, status] =
+                !task || !replacingTeammate || !askingTeammate
+                    ? ['Task replacement successful', 200]
+                    : ['Task not found', 404];
```
hchiam: The `askingTeammate` is asking to give you to take over their task. So is this ternary backwards? Or are you sending a 200 success status if the task and user roles _can't_ be found because that means the id's haven't been swapped yet? Just checking.

server/routes/users.ts
```diff
@@ -144,44 +100,58 @@ router.post('/users', async (req: Request, res: Response, next) => {
...
-            });
-            db.Token.create({
-                userId: addedUser.id,
-                token: token,
-            });
+            }));
 
-            helper.sendVerEmail(username, req, token, 'confirmation');
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+        addedUser && (await db.Token.create({ userId: addedUser.id, token }));
```
hchiam: I got this to pass `eslint server/routes/users.ts` in your branch:
```suggestion
        if (addedUser) await db.Token.create({ userId: addedUser.id, token });
```

hchiam (changes requested):

hchiam (changes requested):
Done reviewing what I can find. Note that there's a merge conflict in [server/routes/tasks.ts -> `router.post('/tasks'`](https://github.com/House-on-a-Rock/tjc-scheduling-server/pull/19/conflicts)

scripts/makeDatabase.js
showduhtung: yeah.. some of these are real weird. I kept the .map and changed up eslintrc instead.

yeah you can disregard a lot of the stuff in this file. it'll never really be used for development, and theyre kinda just personal scripts im writing to help me generate files and data for seed files

server/routes/authentication.ts
showduhtung: yeah. I find it a lot cleaner to declare the constants on the spot, instead of through a huge mumbo jumbo of 'if else' statements

server/db/models/User.ts
showduhtung: kinda. the way I see it, I see Role more as 'team', but.. people don't want me to change role to teams. So every team has many users, and every user is a part of many teams. The inbetween is that UserRole is essentially "Team member" that describes both the user and the team associated with them.

Before, task belonged to both a user and a role, and I thought this doesn't really make much sense when those data points already exists as UserRole. So I'm thinking that we can instead think that the task belongs to a particular "teammate."

server/routes/requests.ts
showduhtung: haha very interesting, I changed it from requestee/requester to your/my because I thought it was clearer (also less wordy and thus cleaner), but as I'm hearing two voices say that requestee makes more sense, then I think that's the way to go.

server/routes/requests.ts
showduhtung: correct.

server/routes/requests.ts
showduhtung: so the `approve` attribute comes from church council. there're certain functions that cannot be swapped without church council approval (sermon speaking). 

but now that I think about it, I didn't really think much about the actual use cases of most of these routes. This pull request/ branch was dedicated to refactoring the code to be cleaner, leaner and just to check for bugs. 

So I think the answer to this is that we don't need to test for requestAll, maybe.. I'm ok with keeping this question in, just as a reminder for the future.

server/routes/tasks.ts
showduhtung: Oops good call, I should've done `task && replacingTeammate && askingTeammate`

server/routes/users.ts
showduhtung: haha thanks for finding these.

PR #22, opened 2020-10-13
St/schedule data

showduhtung:


scripts/makeDatabase.js
```diff
@@ -1,20 +1,26 @@
+/* eslint-disable array-callback-return */
```
showduhtung, line 1: for the most part we can skip this file as having perfect code in this file is not necessary. it's just a few scripts I wrote up to help me generate large files to seed the database.

scripts/seed.js
```diff
@@ -18,34 +20,26 @@ async function seed() {
...
-    await new Promise((r) => setTimeout(r, 2000));
+    const seedChurches = await db.Church.bulkCreate(churches);
+    const seedSchedules = await db.Schedule.bulkCreate(schedules);
+    const seedDividers = await db.Divider.bulkCreate(dividers);
 
-    const seedUsers = await Promise.all(users.map((user) => db.User.create(user)));
-    await new Promise((r) => setTimeout(r, 2000));
-
-    const seedRoles = await Promise.all(roles.map((role) => db.Role.create(role)));
-    await new Promise((r) => setTimeout(r, 2000));
```
hchiam, line 28: nice! just using sequelize bulkCreate

shared/SequelizeTypings/models/EventModel.ts
```diff
@@ -1,9 +1,12 @@
...
+import { CommonSequelizeAttributes } from '.';
 
-export interface EventAttributes {
-    id?: number;
-    time: Date;
-    tag: string;
+export interface EventAttributes extends CommonSequelizeAttributes {
+    roleId?: number; // team
+    day: string;
+    order: number;
```
hchiam, line 7: i'm assuming `order` is just a number to indicate which events happen earlier relative to other events? sounds like an alternative proxy for time/date

shared/SequelizeTypings/models/RequestModel.ts
```diff
@@ -14,8 +14,6 @@ export interface RequestAttributes {
     replace?: boolean;
```
hchiam, line 14: Nice! So someone can ask to give you their task, or assign(?) a task to you, not just swap tasks with you.

scripts/seed.js
showduhtung: right, I totally forgot this existed oops. the previous implementation was creating instances out of order and that was a problem

shared/SequelizeTypings/models/EventModel.ts
showduhtung: not exactly. it's an indicator for how it should be ordered on the schedule. for example

10:45 AM    ||   Hymn Leading
                    ||   Piano
                    ||   Usher

On the schedule, you can tell that Hymn leading, piano and usher are all at 10:45AM, but not all of them will display their time on the row. You also don't want your rows to be out of order accidentally.

shared/SequelizeTypings/models/RequestModel.ts
showduhtung: I think so, right @alan-mj-lin ?

hchiam (approved):
LGTM! :+1:

shared/SequelizeTypings/models/RequestModel.ts
alan-mj-lin: I believe that is the intention yes

hchiam (approved):

hchiam (approved):
LGTM! :+1:

PR #25, opened 2020-11-04
Schedule routes

showduhtung:
adds the first schedule api endpoint. also creates initial schedule models


showduhtung:
there's really only two files to look at:
`server/db/models/Services.ts`
and 
`server/routes/schedules.ts`

server/utilities/helperFunctions.ts
```diff
@@ -97,7 +97,7 @@ export function createToken(tokenType, userId, expiresInMinutes) {
         {
             iss: process.env.AUDIENCE,
             sub: `tjc-scheduling|${userId}`,
-            exp: Math.floor(Date.now() / 1000) + expiresInMinutes * 60,
+            exp: Math.floor(Date.now() / 1000) + expiresInMinutes * 60 * 60,
```
hchiam, line 100: Remove the second `* 60`
`Date.now` is in ms / 1000 = in seconds
expiresInMinutes is in minutes * 60 = seconds, don't need the * 60 again.
seconds + (?) != seconds

server/utilities/helperFunctions.ts
```diff
@@ -152,3 +152,122 @@ export function determineLoginId(auth) {
     const decodedToken = jwt.decode(auth, { json: true });
     return parseInt(decodedToken.sub.split('|')[1], 10);
 }
+
+export function timeToMilliSeconds(time: string) {
+    const [hourMin, period] = time.split(' ');
+    const [hour, min] = hourMin.split(':');
+    const convertedHour = hour === '12' ? 3600000 : 3600000 * parseInt(hour, 10);
```
hchiam, line 159: I think you mean this?
```suggestion
    const convertedHour = hour === '12' ? 0 : 3600000 * parseInt(hour, 10);
```

server/utilities/helperFunctions.ts
```diff
@@ -152,3 +152,122 @@ export function determineLoginId(auth) {
...
+        arr.unshift(target);
+        return arr;
+    }
+    const condition =
+        type === 'order' ? target.order < arr[lastIdx].order : !isTimeBefore(arr[lastIdx].time, target.time);
+    if (!condition) {
+        arr.splice(lastIdx + 1, 0, target);
+        return arr;
+    }
+    if (condition) return correctOrder(arr, lastIdx - 1, target, type);
```
hchiam: You might not need the eslint-disable if you combine the exclusive conditions:
```suggestion
    if (!condition) {
        arr.splice(lastIdx + 1, 0, target);
        return arr;
    } else {
        return correctOrder(arr, lastIdx - 1, target, type);
    }
```

server/utilities/helperFunctions.ts
```diff
@@ -152,3 +152,122 @@ export function determineLoginId(auth) {
...
+    Monday: 1,
+    Tuesday: 2,
+    Wednesday: 3,
+    Thursday: 4,
+    Friday: 5,
+    Saturday: 6,
+};
+
+const zeroPaddingDates = (monthIdx: number, dayIdx: number): string => {
+    let month = (monthIdx + 1).toString();
```
hchiam, line 205: This is nitpicky and not urgent, but I think the `+ 1` to convert the index shouldn't be `zeroPaddingDates`'s responsibility

server/db/models/Services.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ServiceInstance, ServiceAttributes } from 'shared/SequelizeTypings/models/ServiceModel';
+
+const ServiceFactory = (
+    sequelize: Sequelize.Sequelize,
+    DataTypes: Sequelize.DataTypes,
+): Sequelize.Model<ServiceInstance, ServiceAttributes> => {
+    const attributes: SequelizeAttributes<ServiceAttributes> = {
+        name: {
+            type: DataTypes.STRING,
+            allowNull: true,
```
hchiam: Looks like you simply renamed the file, but should it be `allowNull: false,` here? I think [sequelize defaults `allowNull` (nullable) to true](https://sequelize.org/master/manual/model-basics.html?:~:text=Using%20sequelize.define%3A#using--a-href-----class-lib-sequelize-js-sequelize-html-instance-method-define---code-sequelize-define--code---a--:~:text=Using%20sequelize.define%3A).

server/routes/schedules.ts
```diff
@@ -0,0 +1,83 @@
...
+        const searchParams = { [Op.and]: searchArray };
+        const servicesData = await db.Service.findAll({ where: searchParams });
+
+        const schedulesData = await Promise.all(
+            schedules.map(
+                async ({ id: scheduleId, title: scheduleTitle, view, start: scheduleStart, end: scheduleEnd }) => {
+                    const eventsData = await db.Event.findAll({ where: { scheduleId } });
+                    const services = await Promise.all(
+                        servicesData.map(async (service) => {
+                            const { name, start: startofService, end: endOfService, day: dayOfService } = service;
```
hchiam: minor: camelCase on your alias `startofService` -> `startOfService`

server/routes/schedules.ts
```diff
@@ -0,0 +1,83 @@
...
+                                        );
+
+                                        // need to add typescript types to these data structures
+                                        const serviceData: any = { duty: { data: { display: title } } };
+                                        if (order === 1) serviceData.time = { data: { display: time } };
+                                        tasksData.map((task) => {
+                                            const key = Object.keys(task)[0];
+                                            const value = Object.values(task)[0];
+                                            serviceData[key] = value;
+                                        });
```
hchiam: why are we mapping to get serviceData have the first (`[0]`) kvp of each task? I think `Object.keys()` and `Object.values()` don't guarantee order (not sure about the DB call though)

hchiam (changes requested):

server/utilities/helperFunctions.ts
showduhtung: haha tbh this is ok because this is for development purposes. The previous expiration token expired too fast when I'm testing things, and I'd rather a longer token while developing

server/routes/schedules.ts
showduhtung: so the reason why was because 
```
const tasksData = [{
  [contrivedDate(date)]: {
    data: { firstName, lastName, userId, role }
  }
}]
```
which meant that there'd be only one key and one value for each, but looking at this over, I realized that trying to organize the data in the tasksData makes no sense, when it can be done when mapping through tasksData to construct the serviceData


so I'll be fixing this to be more efficient

server/utilities/helperFunctions.ts
showduhtung: hmm, but then you'd have to remember whenever you put them in. trade offs either way, but I think this looks cleaner. do you have another suggestion other than inputting `zeroPaddingDates(monthIdx + 1, dayIdx)`?

server/utilities/helperFunctions.ts
showduhtung: agreed. one thing about your suggestion is that airbnb eslint (which is industry standard) doesn't like 
`else return` saying that else is redundant

so I'll be making it
```
  if (!condition) {
        arr.splice(lastIdx + 1, 0, target);
        return arr;
    }
    return correctOrder(arr, lastIdx - 1, target, type);
```

server/utilities/helperFunctions.ts
hchiam: expiresInMinutes isn't really minutes anymore but it sounds like it prolly won't be too serious then. i'm guessing you're just changing it in one spot here, instead of doing a global edit of createToken uses

server/utilities/helperFunctions.ts
hchiam: as long as we remember which way it is for months it should be fine. otherwise you have to read the function, but it's not that bad in this case.

server/utilities/helperFunctions.ts
hchiam: i marked everything else as resolved. just this one question left

hchiam (changes requested):
i marked everything as resolved except for the `0 : 3600000 * parseInt(hour, 10);`

server/utilities/helperFunctions.ts
hchiam: yup sounds right

PR #28, opened 2020-11-11
St/seed nov

showduhtung:


server/routes/authentication.ts
```diff
@@ -63,6 +63,67 @@ router.post('/resendConfirm', async (req: Request, res: Response, next: NextFunc
...
+                default:
+                    return ['success', 200];
+            }
+        };
+        const [message, status] = determineMessageStatus();
+
+        if (status > 299) return res.status(status).send({ message });
+        if (hashedLoginPassword !== password) {
+            const updatedAttempts =
+                loginAttempts === 3
```
hchiam: I'd write `>= 3` just to be safe, instead of only checking iff it's exactly 3. Especially if there's already something fishy going on with too many attempts.

hchiam (changes requested):

server/routes/authentication.ts
showduhtung: sounds good

PR #30, opened 2020-11-14
Tc/table fix

CedThen:


server/routes/schedules.ts
```diff
@@ -73,7 +79,7 @@ router.get('/schedules', certify, async (req: Request, res: Response, next: Next
 
 router.post('/schedules', certify, async (req: Request, res: Response, next: NextFunction) => {
     try {
-        const { title, view, startDate, endDate, churchId, roleId } = req.body;
+        const { title, view, startDate, endDate, churchId, team } = req.body; //roleId still team on webapp
```
showduhtung, line 82: ok but we're passing an 'id' through and not something like a 'team name'. We can talk about renaming roles to teams?

showduhtung (approved):
looks good to me. I'll want to continue thinking on the endpoint naming for `/schedules/tabs` and `/schedules`, and how `*/tabs` is currently coming before `*~/` but otherwise lgtm. change the team back to roleId for now, just because it's still an id. If we want to change roleId --> teamId, we should do a sweeping change for it

PR #31, opened 2020-12-19
Tc/cell mutation

CedThen:


server/db/models/Task.ts
```diff
@@ -9,13 +9,14 @@ const TaskFactory = (
...
         status: { type: DataTypes.STRING, defaultValue: 'active' },
+        // id: { type: DataTypes.INTEGER, primaryKey: true },
     };
 
     const Task = sequelize.define<TaskInstance, TaskAttributes>('Task', attributes);
 
     Task.associate = (models) => {
         Task.belongsTo(models.Event, { as: 'event', foreignKey: 'eventId' });
-        Task.belongsTo(models.UserRole, { as: 'user', foreignKey: 'userId' });
+        Task.belongsTo(models.User, { as: 'user', foreignKey: 'userId' });
```
showduhtung: does work in the seed scripts?

server/db/models/User.ts
```diff
@@ -91,6 +91,7 @@ const UserFactory = (
 
     User.associate = (models) => {
         User.belongsTo(models.Church, { as: 'church', foreignKey: 'churchId' });
+        // User.hasMany(models.Tasks, { as: 'task', foreignKey: 'taskId' });
```
showduhtung: you can delete this

server/db/models/Task.ts
CedThen: yes

server/routes/schedules.ts
```diff
@@ -26,48 +26,22 @@ router.get('/schedules', certify, async (req: Request, res: Response, next: Next
         const { scheduleId } = req.query;
         const schedule = await db.Schedule.findOne({ where: { id: scheduleId.toString() } });
         const scheduleRole = await db.Role.findOne({ where: { id: schedule.roleId } });
```
showduhtung: did I name this? we should probably not name it scheduleRole, just role. scheduleRole is a little redundant.

server/routes/tasks.ts
```diff
@@ -90,37 +90,23 @@ router.delete('/tasks/:taskId', certify, async (req: Request, res: Response, nex
...
-                await switchTask.update({
-                    id: switchTask.id,
-                    userId: targetTask.userId,
-                });
+            if (targetTask && targetTask.userId !== parseInt(assigneeId)) {
+                await targetTask.update({ userId: assigneeId });
+                return res.status(200).send('Tasks updated');
             }
-            return res.status(status).send({ message });
+            return res.status(400).send({ message: 'error' });
```
showduhtung: this is ok for now since I'm gonna add this back in at a later date, but if you could use the const [message, status] method to be consistent for all apis

showduhtung (changes requested):
can you work on making sure the new ```Task.belongsTo(models.User ...``` also works with the seeding script

server/db/models/Task.ts
showduhtung: but you didn't make changes to the seeding scripts to change userRoleId --> userId

CedThen: they do say userId tho

server/routes/schedules.ts
CedThen: yeah i think u named that haha

CedThen: renamed

server/routes/tasks.ts
CedThen: ok for now is good enough for me

PR #35, opened 2020-12-23
TSA-124 delete schedule functionality 

alan-mj-lin:
- added cascade deletes to Service, Event, and Schedule models
- button press deletes schedule and all associated items
- handles case for when there are no schedules


src/components/Main/Schedule/Home.tsx
```diff
@@ -75,6 +89,10 @@ export const Home = () => {
     });
   }
 
+  async function onScheduleDelete(scheduleId: string, title: string) {
```
showduhtung, line 88: so you don't actually need to have this async-await if all this does is mutateDeleteSchedule

you can write:
```
function onScheduleDelete(scheduleId: string, title: string) {
 mutateDeleteSchedule({ scheduleId, title });
}
```

showduhtung: async await is only really necessary if you need to wait for the mutateDeleteSchedule to finish before going onto the next step within the scope of the same function

so if you were to do:
```
async function onScheduleDelete(scheduleId: string, title: string) {
   const response = await mutateDeleteSchedule({ scheduleId, title });
   console.log(response)
}
```
This would be legit. Whereas this would console.log a Promise
```
function onScheduleDelete(scheduleId: string, title: string) {
   const response = mutateDeleteSchedule({ scheduleId, title });
   console.log(response)
}
```

src/components/Main/Schedule/Home.tsx
```diff
@@ -86,6 +104,15 @@ export const Home = () => {
       >
         Log Out
       </Button>
+      <Button
+        onClick={() => {
+          setOpenedTabs([0]);
```
showduhtung, line 96: does this fix the hover issue?

src/components/Main/Schedule/Home.tsx
```diff
@@ -94,7 +121,7 @@ export const Home = () => {
         />
       </Dialog>
       {alert && <Alert alert={alert} unMountAlert={() => setAlert(null)} />}
-      {data && (
+      {data && data.length > 0 && (
```
showduhtung, line 111: we can change this to
```
data?.length > 0 && (...
```
these are the same thing

server/db/models/Event.ts
```diff
@@ -23,7 +23,11 @@ const EventFactory = (
   const Event = sequelize.define<EventInstance, EventAttributes>('Event', attributes);
 
   Event.associate = (models) => {
-    Event.belongsTo(models.Service, { foreignKey: 'serviceId' });
+    Event.belongsTo(models.Service, {
+      foreignKey: 'serviceId',
+      onDelete: 'cascade',
+      hooks: true,
```
showduhtung, line 29: what does `hooks: true` do?

showduhtung (commented):
Is this a fully functioning pr?

server/db/models/Services.ts
```diff
@@ -26,8 +26,16 @@ const ServiceFactory = (
   );
 
   Service.associate = (models) => {
-    Service.belongsTo(models.Schedule, { as: 'schedule', foreignKey: 'scheduleId' });
-    // Service.hasMany(models.Event);
+    Service.belongsTo(models.Schedule, {
+      as: 'schedule',
+      foreignKey: 'scheduleId',
+    });
+    Service.hasMany(models.Event, {
```
CedThen, line 33: I think we left this off because in the Event factory, we have events belonging to service, thus this wasn't required. Shaun can you confirm?

same for the schedule.hasMany services

src/components/Main/Schedule/Home.tsx
CedThen: looks like its just a button to trigger deleting the current tab

src/components/Main/Schedule/Home.tsx
```diff
@@ -25,18 +25,32 @@ export const Home = () => {
...
       cache.invalidateQueries('scheduleTabs');
       closeDialogHandler(response);
     },
   });
 
+  const [mutateDeleteSchedule, { error: deleteScheduleError }] = useMutation(
+    deleteSchedule,
+    {
+      onSuccess: (response) => {
+        console.log(response);
```
CedThen, line 44: Can you remove the console.log's that you added? Also means u can turn this function into a one-liner without the curly braces

server/routes/schedules.ts
```diff
@@ -98,3 +98,25 @@ router.post(
...
   },
 );
+
+router.delete(
+  '/schedules',
+  certify,
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const { scheduleId, title } = req.body;
+      console.log(scheduleId, title);
```
CedThen, line 108: remove console logs

CedThen:
Please also use the Alert component to have the notification bar pop-up to show that the action was completed by the server.

server/db/models/Services.ts
showduhtung: So I read up on sequelize hasMany and belongsTo and why we need to use it. https://stackoverflow.com/questions/45389747/sequelize-hasmany-belongsto-or-both

So it looks like we do need to have hasMany if we want to do a cascade delete, which we do. HasMany relationship specifies all the related data that needs to be deleted when the schedule is deleted.

src/components/Main/Schedule/Home.tsx
alan-mj-lin: if i don't have that there i get an error after deleting

server/db/models/Event.ts
alan-mj-lin: if i dont do hooks true, the onDelete: 'cascade' doesn't happen. Meaning I get foreignkey error when trying to delete a schedule with relationships to Services and Events

src/components/Main/Schedule/Home.tsx
alan-mj-lin: > Is this a fully functioning pr?

yes this is a fully functional pr

PR #37, opened 2020-12-25
TSA-124 & TSA-132 delete button functionality with alerts

alan-mj-lin:
- addressed all comments from pull request #35 
- alert now uses proper material ui snackbar library
- alerts show on adding a schedule and deleting a schedule
- alerts should show on server errors as well (not tested)

### DIFF
--- a/src/components/Schedule/ScheduleContainer.tsx
+++ b/src/components/Schedule/ScheduleContainer.tsx
@@ -40,10 +40,7 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
   const [mutateDeleteSchedule, { error: deleteScheduleError }] = useMutation(
     deleteSchedule,
     {
-      onSuccess: (response) => {
-        console.log(response);
-        cache.invalidateQueries('scheduleTabs');
-      },
+      onSuccess: () => cache.invalidateQueries('scheduleTabs'),
     },
   );
 
@@ -85,10 +82,9 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
     });
   }
 
-  async function onScheduleDelete(scheduleId: string, title: string) {
-    await mutateDeleteSchedule({ scheduleId, title });
+  function onScheduleDelete(scheduleId: string, title: string) {
+    mutateDeleteSchedule({ scheduleId, title });
   }
-  console.log(openedTabs);
   return (
     <>
       <Button
@@ -108,7 +104,7 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
         />
       </Dialog>
       {alert && <Alert alert={alert} unMountAlert={() => setAlert(null)} />}
-      {data && data.length > 0 && (
+      {data.length > 0 && (
         <div>
           <ScheduleTabs
             tabIdx={tabIdx}
--- a/src/components/Schedule/ScheduleContainer.tsx
+++ b/src/components/Schedule/ScheduleContainer.tsx
@@ -9,7 +9,7 @@ import { addSchedule, deleteSchedule } from '../../query/apis/schedules';
 import { ScheduleTabs } from './ScheduleTabs';
 import { NewScheduleForm } from './NewScheduleForm';
 import { ScheduleComponent } from './ScheduleComponent';
-import { CustomSnackbar } from '../shared/Alert';
+import { Alert } from '../shared/Alert';
 import { ConfirmationDialog } from '../shared/ConfirmationDialog';
 import { AlertInterface } from '../../shared/types';
 import { buttonTheme } from '../../shared/styles/theme';
@@ -42,10 +42,8 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
     deleteSchedule,
     {
       onSuccess: (response) => {
-        console.log(response.data);
         cache.invalidateQueries('scheduleTabs');
         setAlert({ message: response.data, status: 'success' });
-        console.log(alert !== null);
       },
     },
   );
@@ -89,9 +87,6 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
     });
   }
 
-  function onScheduleDelete(scheduleId: string, title: string) {
-    mutateDeleteSchedule({ scheduleId, title });
-  }
   return (
     <>
       <Button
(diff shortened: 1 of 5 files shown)


src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -41,13 +42,16 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
     deleteSchedule,
     {
       onSuccess: (response) => {
-        console.log(response);
+        console.log(response.data);
```
hchiam: Remember to remove the [console log](https://memegenerator.net/img/instances/68620658/a-wild-consolelog-appeared.jpg) if it's not needed.

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -41,13 +42,16 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
...
         cache.invalidateQueries('scheduleTabs');
+        setAlert({ message: response.data, status: 'success' });
+        console.log(alert !== null);
```
hchiam: Remember to remove the [console log](https://memegenerator.net/img/instances/68620658/a-wild-consolelog-appeared.jpg) if it's not needed.

hchiam (changes requested):
A couple minor things with the `console.log`s.

We could have a better system for the `z-index`es, but I'd leave it for now since it's easy to clean up later if we really need to.

src/components/shared/Alert.tsx
```diff
@@ -46,9 +38,8 @@ const useStyles = makeStyles((theme: Theme) =>
       top: 20,
       left: 20,
       right: 20,
-      zIndex: 100000, // too much?
+      zIndex: 1000000, // too much?
```
hchiam: A little unusual. I think this is greater than the loading spinner's `z-index`, but shouldn't be a problem. If we really want to, we can make it more semantic and have named `zIndex`es in theme.js, but I think it's fine to defer this to when we really need it.

alan-mj-lin: Alert isn't currently using that set of styles anyway haha

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -85,17 +89,13 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
...
   }
-  console.log(openedTabs);
   return (
     <>
       <Button
-        onClick={() => {
-          setOpenedTabs([0]);
-          onScheduleDelete(data[tabIdx].id, data[tabIdx].title);
-        }}
+        onClick={() => setDeleteDialogOpen(!isDeleteDialogOpen)}
```
CedThen, line 93: This isn't sticky to the tabs bar

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -107,7 +107,24 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
           onSubmit={onNewScheduleSubmit}
         />
       </Dialog>
-      {alert && <Alert alert={alert} unMountAlert={() => setAlert(null)} />}
+      <ConfirmationDialog
+        title="Are you sure you want to delete?"
+        isOpen={isDeleteDialogOpen}
+        handleClick={(yesOrNo) => {
+          if (yesOrNo === true) {
```
CedThen: clicking no doesn't do anything?
try using a ternary statement when u include the no option functionality

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -107,7 +107,24 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
...
+            setOpenedTabs([0]);
```
CedThen, line 111: when there are no tabs left, schedule tabs seems to unmount. Can you make sure that it stays mounted, so that + button is still functional?

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -107,7 +107,24 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
...
+            onScheduleDelete(data[tabIdx].id, data[tabIdx].title);
```
CedThen: You don't really need to make a function to just do this. Im guessing you followed onNewScheduleSubmit, can you also rewrite it so the await part is included inside the useMutation, and get rid of onNewScheduleSubmit?

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -107,7 +107,24 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
...
+            setDeleteDialogOpen(!isDeleteDialogOpen);
```
CedThen: I think this should be name setIsDeleteDialogOpen
Maybe just set this to false since you're closing it, can potentially avoid bugs? @showduhtung whats the typical pattern here?

src/components/shared/Alert.tsx
```diff
@@ -1,38 +1,30 @@
...
-
-  useEffect(() => {
-    setTimeout(unMountAlert, displayTimer); // unmount after 2 seconds
-  });
-  // only success has been implemented, since error is handled inside each form. But we'll pbly run into use cases for the error alert
-  const muiAlertProps: { icon: JSX.Element; severity: Color } =
-    alert.status === 'success'
-      ? { severity: 'success', icon: <CheckIcon fontSize="inherit" /> }
-      : { severity: 'error', icon: <ErrorIcon fontSize="inherit" /> };
+function Alert(props: AlertProps) {
```
CedThen: this function is unnecessary, I'd prefer if you removed this function and renamed the export back to alert

src/components/shared/Alert.tsx
```diff
@@ -1,38 +1,30 @@
...
-      icon={muiAlertProps.icon}
-      severity={muiAlertProps.severity}
-      className={classes.root}
-      variant="filled"
-    >
-      {alert.message}
-    </MuIAlert>
+    <Snackbar open={isOpen} autoHideDuration={2000} onClose={handleClose}>
+      <Alert
+        severity={alert.status === 'success' ? 'success' : 'error'}
```
CedThen: You can type this to only be 'success' or 'error', and then just pass alert.status to severity instead of this ternary

src/components/Schedule/ScheduleContainer.tsx
alan-mj-lin: im not sure what you mean by putting await inside useMutation

hchiam (commented):
Leaving it to Ted/Shaun. Not sure if a couple of Ted's comments are resolved yet.

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -111,36 +106,37 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
         title="Are you sure you want to delete?"
         isOpen={isDeleteDialogOpen}
         handleClick={(yesOrNo) => {
+          setDeleteDialogOpen(!isDeleteDialogOpen);
           if (yesOrNo === true) {
```
CedThen: You can just do if (yesOrNo) 
I'm fine with merging this

src/components/Schedule/ScheduleContainer.tsx
```diff
@@ -107,23 +102,41 @@ export const ScheduleContainer = ({ churchId }: ScheduleContainerProps) => {
...
         />
       </Dialog>
-      {alert && <Alert alert={alert} unMountAlert={() => setAlert(null)} />}
-      {data && data.length > 0 && (
+      <ConfirmationDialog
+        title="Are you sure you want to delete?"
+        isOpen={isDeleteDialogOpen}
+        handleClick={(yesOrNo) => {
+          setDeleteDialogOpen(!isDeleteDialogOpen);
+          if (yesOrNo) {
```
showduhtung: maybe instead of yesOrNo, we can go with open. 

because !open = close, but !yesOrNo = yesOrNo

alan-mj-lin: but yesOrNo doesn't correspond to whether the dialogue box is open or not. It corresponds to whether the user clicked yes or no on the dialogue. Yeah I agree yesOrNo isn't the best of variable names lol but it's just what I did in the moment hahaha

showduhtung: ohohoh. then clicked works right? since right now it's arbitrary what the user clicked. 

so yesOrNo = clicked, and !yesOrNo = not clicked? correct me if I'm wrong
so then you could just use click, !click lol

whatever the case, since I'm assuming it's gonna be changed anyways it doesn't matter, so you have my approval to merge

alan-mj-lin: yesOrNo = true means the user clicked Yes. yesOrNo= false means the user clicked no.

PR #38, opened 2020-12-29
delete event functionality with confirmation dialog

alan-mj-lin:
idk if theres a JIRA ticket for this but tag it here if there is somebody.
- delete event functionality on right-clicking delete row
- confirmation dialogue


showduhtung (approved):
There's a bunch of things in here that conflict with my branch. So I'll make the changes in my branch once we merge this. 

Just uhm.. let's name clickedYes --> accepted, as in the user has accepted the risk of deleting a schedule. 

But otherwise, lgtm

PR #39, opened 2020-12-30
St/restructure

showduhtung:


src/components/Auth/AuthProvider.tsx
```diff
@@ -2,19 +2,14 @@ import React, { useState } from 'react';
...
 export const AuthProvider = ({ children }: any) => {
-  const [isLoggedIn, setIsLoggedIn] = useState(false);
-  // const [token, setToken] = React.useState(null);
-  const login = () => {
-    setIsLoggedIn(true);
-  };
-  const logout = () => {
-    setIsLoggedIn(false);
-  };
+  const [isLoggedIn, setIsLoggedIn] = useState(true);
```
alan-mj-lin, line 5: Are we still working on a solution to this problem?

src/components/Auth/PrivateRoute.tsx
```diff
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import { Route, Redirect } from 'react-router-dom';
+import { AuthContext } from '../../shared/services/AuthContext';
 import { extractTokenInfo, useToken } from '../../shared/utilities';
 
 const validateUser = (condition: string, token): boolean => {
```
alan-mj-lin, line 6: access and exp in validateUser are unused. Idk if it's being used in the future but if not probably good to get rid of it.

src/components/Schedule/Schedule.tsx
```diff
@@ -0,0 +1,270 @@
...
+  // Save Data
+  function onTaskModified(taskId: number, newAssignee: number, isChanged: boolean) {
+    if (isChanged) {
+      const updatedChangedTasks = { ...changedTasks.current, [taskId]: newAssignee };
+      changedTasks.current = updatedChangedTasks;
+    } else if (changedTasks.current[taskId]) delete changedTasks.current[taskId];
+
+    setIsScheduleModified(Object.keys(changedTasks.current).length > 0);
+  }
+  function onSaveScheduleChanges() {
```
alan-mj-lin, line 94: do you really need to write this as a separate function?

src/components/Schedule/Schedule.tsx
```diff
@@ -0,0 +1,270 @@
...
+            tabs={tabs}
+            handleAddClicked={() => setIsNewScheduleOpen(true)}
+          />
+          <ScheduleToolbar
+            handleNewServiceClicked={() => setIsNewServiceOpen(true)}
+            destroySchedule={() =>
+              removeSchedule({ scheduleId: tabs[tab].id, title: tabs[tab].title })
+            }
+            isScheduleModified={isScheduleModified}
+            onSaveScheduleChanges={onSaveScheduleChanges}
```
alan-mj-lin, line 174: couldn't you just do onSaveScheduleChanges={setIsScheduleModified(false)}

src/components/Schedule/ScheduleTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+import MaUTable from '@material-ui/core/Table';
+import TableHead from '@material-ui/core/TableHead';
+import TableBody from '@material-ui/core/TableBody';
+import TableRow from '@material-ui/core/TableRow';
+import { makeStyles, Theme, createStyles, fade, darken } from '@material-ui/core/styles';
+
+// Styles
+import { paletteTheme } from '../../shared/styles/theme.js';
+
+export const ScheduleTable = ({ title, hidden, children }: any) => {
```
alan-mj-lin: eventually add proper typing for props?

src/components/Schedule/ScheduleTableBody.tsx
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable react/no-array-index-key */
+import React from 'react';
+import TableRow from '@material-ui/core/TableRow';
+import TableCell from '@material-ui/core/TableCell';
+import { makeStyles, Theme, createStyles } from '@material-ui/core/styles';
+
+export const ScheduleTableBody = ({ title, children }: any) => {
+  //   const classes = useStyles();
```
alan-mj-lin, line 8: a lot of dead code

src/components/Schedule/ScheduleTableHeader.tsx
```diff
@@ -0,0 +1,29 @@
...
+import React from 'react';
+
+// Material UI
+import TableCell from '@material-ui/core/TableCell';
+import { makeStyles, Theme, createStyles } from '@material-ui/core/styles';
+
+// Styles
+import { typographyTheme } from '../../shared/styles/theme.js';
+
+export const ScheduleTableHeader = ({ header }: any) => {
```
alan-mj-lin, line 11: proper typings for props

src/components/Schedule/ScheduleToolbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+    </Tooltip>
+  );
+};
+
+export const ScheduleToolbar = ({
+  handleNewServiceClicked,
+  destroySchedule,
+  isScheduleModified,
+  onSaveScheduleChanges,
+}: any) => {
```
alan-mj-lin: proper typings for props

src/components/Schedule/TableCell.tsx
```diff
@@ -27,22 +27,34 @@ export const DataCell = React.memo(
       setValue(newValue);
     }
 
-    const cellStyle = isCellModified ? classes.modified : classes.cell; // normally i'd combine styling objects to reduce repetition, but these are strings so idk
-
+    function renderOptions(option: any, state: any) {
```
alan-mj-lin, line 30: I thought typing as any was bad practice

src/query/apis/schedules.ts
```diff
@@ -1,53 +1,29 @@
...
+  axios.post(`${secretIp}/api/schedule`, data);
 
-export const deleteSchedule = ({ scheduleId, title }: DeleteScheduleProps) =>
-  axios.delete(`${secretIp}/api/schedules`, {
-    data: {
-      scheduleId,
-      title,
-    },
-  });
+export const destroySchedule = (data: DeleteScheduleProps) =>
```
alan-mj-lin: I think I like this

src/query/schedules.ts
```diff
@@ -1,7 +1,19 @@
...
+    : [];
+};
 
-export const getTabData = async (key: string, churchId: number) =>
-  (await getTabs(churchId)).data;
+export const getAllSchedules = async (churchId: number) => {
+  return (await getSchedules(churchId)).data;
+};
+
+// export const createNewSchedule = async (newInfo: NewScheduleData) =>
```
alan-mj-lin, line 17: dead code

src/query/services.ts
```diff
@@ -0,0 +1,5 @@
+export {}; // import { NewServiceData } from './apis/services';
```
alan-mj-lin, line 1: what's with the empty export?

src/shared/types/models.ts
```diff
@@ -142,17 +135,24 @@ export interface AddUserProps {
...
+export interface NewServiceData {
+  name: string;
+  order?: number;
+  dayOfWeek: number;
+  scheduleId?: number;
 }
 
 export interface DeleteScheduleProps {
-  scheduleId: string;
+  scheduleId: number;
```
alan-mj-lin, line 156: I originally typed it as string because when you get the json object that id is of a string type. Idk how using number affects the call.

alan-mj-lin (changes requested):
See comments

package.json
```diff
@@ -89,7 +90,7 @@
     "react-contextmenu": "^2.14.0",
     "react-dom": "^16.9.0",
     "react-navigation": "^4.3.9",
-    "react-query": "^2.26.2",
+    "react-query": "^3.5.0",
```
hchiam, line 93: Why was package-lock.json removed? The `/src` folder has its own lock file and package.json.

src/components/Schedule/NewScheduleForm.tsx
```diff
@@ -77,9 +77,13 @@ export const NewScheduleForm = ({ onClose, error, onSubmit }: NewScheduleFormPro
 
   return (
     <div className={classes.root}>
-      <h2>New Schedule Form</h2>
+      <h2>Create a new schedule</h2>
```
hchiam:
```suggestion
      <h2>Create A New Schedule</h2>
```
Capitalize the `<h2>`, to be consistent with NewServiceForm.js's "Add A New Event"

src/components/Auth/PrivateRoute.tsx
```diff
@@ -10,8 +11,13 @@ const validateUser = (condition: string, token): boolean => {
 };
 
 export const PrivateRoute = ({ children, redirection, condition, ...rest }: any) => {
+  const auth = useContext(AuthContext);
```
hchiam, line 25: `useContext`. nice.

src/components/Schedule/NewScheduleForm.tsx
```diff
@@ -9,18 +9,24 @@ import { buttonTheme } from '../../shared/styles/theme.js';
...
-  error: any;
-  onSubmit: (
-    title: string,
-    startDate: string,
-    endDate: string,
-    view: string,
-    team: number,
-  ) => void;
+  error: HttpErrorProps;
+  onSubmit: (newScheduleData: NewScheduleData) => void;
```
hchiam: much cleaner

src/components/Schedule/Schedule.tsx
```diff
@@ -0,0 +1,270 @@
...
+  addSchedule: (newInfo: NewScheduleData) => void;
+  addService: (newInfo: NewServiceData) => void;
+  removeSchedule: (info: DeleteScheduleProps) => void;
+}
+
+// Task List
+// 1. When the container has more than 3 tabs, when you select the 4th, how to handle data from data handler
+// 2. Need to update changedTask and the buttons to reflect only on the schedule selected
+// 3. Cell data poorly describes the different kinds of cells that exist. The data structure needs a revamp, and so does the Tablecell/Datacell
+// 4. Low priority but finding scheduleId and order is a pain the way I have it currently implemented because allScheduleData exists in tabs, but singular schedule data exist in schedules. Need to consolidate some of the logic together
```
hchiam, line 56: Don't forget to do these later, as separate [task(s)](https://tjc-scheduling.atlassian.net/jira/software/projects/TSA/boards/1?label=Web) or [issue(s)](https://github.com/House-on-a-Rock/tjc-scheduling-app/issues) or [PR(s)](https://github.com/House-on-a-Rock/tjc-scheduling-app/pulls).

src/components/Schedule/Schedule.tsx
```diff
@@ -0,0 +1,270 @@
...
+                      </ScheduleTableBody>
+                    );
+                  })}
+                </ScheduleTable>
+              </div>
+            );
+          })}
+        </>
+      ) : (
+        <CircularProgress />
```
hchiam, line 252: nice! Pure CSS should have better perf than JS modifying the DOM, especially on slow connections. But using CircularProgress is more consistent with how we use Material-UI elsewhere, and easier to use in this case, it looks like.

src/components/Schedule/Schedule.tsx
```diff
@@ -0,0 +1,270 @@
...
+      // setIsNewScheduleOpen(false);
+    }
+  }, [isSuccess]);
+
+  const teammates = (roleId) =>
+    data.users.filter((user: any) => user.teams.some((role: any) => role.id === roleId));
+
+  return (
+    <>
+      {data?.schedules && data?.users && !isLoading ? (
```
hchiam, line 159: for easier reading, i'd "flip" the ternary condition ([de morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws#cite_ref-3:~:text=the%20negation%20of%20a%20conjunction%20is%20the%20disjunction%20of%20the%20negations)) so we can move the shorter alternative output to be directly under the condition, like this:
```suggestion
      {!data?.schedules || !data?.users || isLoading ? (
          <CircularProgress />
      ) : (
          /* the longer, main schedule tabs stuff */
      )}
```
Instead of having the `<CircularProgress />` suddenly at the end.

src/components/Schedule/ScheduleTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+      // first column:
+      '& td:first-child, th:first-child': {
+        left: '8px',
+        width: '12ch', // need this when there's very few columns
+        '& input': {
+          width: '12ch',
+          textAlign: 'center',
+        },
+        '&:before': {
```
hchiam, line 66:
```suggestion
        '&:before': { // use a pseudo-element to cover left-side gap when scrolling
```
I forgot to comment this. When I scrolled items left, I could see items "peeking out" from behind the table on the left side of the table when you scroll. Used a CSS pseudo-element to cover that gap.

src/components/Schedule/ScheduleTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+      // second column:
+      '& td:nth-child(2), th:nth-child(2)': {
+        left: '135px',
+        width: '14ch', // need this when there's very few columns
+        '& input': {
+          width: '14ch',
+        },
+        borderRightWidth: 0,
+        '&:after': {
```
hchiam, line 86:
```suggestion
        '&:after': { // use a pseudo-element to cover right-side gap when scrolling
```
I forgot to comment this. When I scrolled, there was a weird tiny gap that was showing up on the right side of the frozen columns. Used a CSS pseudo-element to cover that other gap.

src/components/Schedule/ScheduleTableBody.tsx
hchiam: @showduhtung are you keeping these here just in case we want to revert the JSX and/or reuse the `useStyles`?

src/components/Schedule/TableCell.tsx
hchiam: prolly WIP

src/components/Schedule/TableCell.tsx
```diff
@@ -27,22 +27,34 @@ export const DataCell = React.memo(
...
+      const { firstName, lastName, userId } = option;
+      // would like this to autosize so each option is just a single line
```
hchiam, line 32: To do this, `width: '100%',` on each option might work. 
If that doesn't work, try making the parent use `display: 'flex'` and each child have `flex: 1`

src/components/Schedule/utilities.ts
```diff
@@ -92,8 +92,8 @@ export function isInTime(target: string, start: string, end: string): boolean {
 export function timeToMilliSeconds(time: string) {
   const [hourMin, period] = time.split(' ');
   const [hour, min] = hourMin.split(':');
-  const convertedHour = hour === '12' ? 3600000 : 3600000 * parseInt(hour);
-  const convertedMin = 60000 * parseInt(min);
+  const convertedHour = hour === '12' ? 3600000 : 3600000 * parseInt(hour, 10);
+  const convertedMin = 60000 * parseInt(min, 10);
```
hchiam, line 96: Good catch. The `parseInt` radix could become octal when the hour/min strings start with 0 ([in older browsers](https://www.w3schools.com/jsref/jsref_parseint.asp#main:~:text=This%20feature%20is%20deprecated) - now [all browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Browser_compatibility:~:text=Parses%20leading%2Dzero%20strings%20as%20decimal%2C%20not%20octal) default to decimal in this edge case).

src/components/Auth/AuthProvider.tsx
showduhtung: nope

src/components/Auth/PrivateRoute.tsx
showduhtung: oh no, access and expiration will definitely need to be used. i rushed to finish this branch so that you and ted can start working on it.

src/components/Schedule/Schedule.tsx
showduhtung: this was ted's original function. in it was a note that said that there is an assumption many other things will change. the future of the changedTask ref will track changes for each tab. the save changes button will reflect it as well.

src/components/Auth/AuthProvider.tsx
alan-mj-lin: do you want me to look into it?

src/query/apis/schedules.ts
hchiam: much cleaner

src/components/Schedule/Schedule.tsx
showduhtung: responded in a previous commentj

src/components/Schedule/ScheduleTable.tsx
showduhtung: oh thanks. yah i'll actually go ahead and do that now because why not

src/components/Schedule/ScheduleTableBody.tsx
showduhtung: actually it's not dead code but my commented out comments that includes code. the comments have notes and thoughts for a feature that ted and I were talking about, and I think it makes sense to add it here. but for now, it's a very reusable and functional component

showduhtung: also, I'd argue that commented out code isn't actually dead, as it's not being read. unused code is dead

src/components/Schedule/TableCell.tsx
showduhtung: it is.. a lot of this was ted's stuff.

src/query/apis/schedules.ts
showduhtung: heh

src/shared/styles/theme.js
```diff
@@ -159,7 +159,7 @@ export const themeExtension = {
       color: 'black',
       fontWeight: 'bold',
     },
-    '&:hover, &:focus': {
+    '&:hover': {
```
hchiam, line 162: why is this removed? having a `&:focus` style makes it more obvious to people using screen readers when they navigate tabs by hitting the `tab` key, if they can't use a touchscreen or a mouse to hover (a11y)

src/query/schedules.ts
showduhtung: im keeping it for now, because there's currently a discrepancy with how i'm using query and apis. I'm using some endpoints directly from api, and then im' using some endpoints through and intermediary function via query. I haven't yet decided how it should be used, but once I make data handlers for the other containers, then a pattern should emerge. until then, this, as well as the `export {}` in services.ts will stay as reference for myself when i continue working on this

src/query/services.ts
showduhtung: typescript won't compile because i have it required that every file must be a module. so empty files get flagged as not a module

src/shared/types/models.ts
showduhtung: it doesn't really. but I've noticed that we're not being consistent with the id's throughout the app. i haven't had the brain power to go through them all and rewrite them as such. but yah maybe we should change so that the app expects the id to be a string. we just have to go through with a fine comb looking for these differences

showduhtung: but if we can do it later in a different branch thatd be great because that's outside the scope of this pull request

hchiam (changes requested):

src/components/Schedule/TableCell.tsx
hchiam: Or leave this to another task/ticket.

showduhtung:
hey @alan-mj-lin @CedThen @hchiam , so just a note of this branch- this branch is NOT finished. there're no data handlers or containers for templates/team/members yet. this pull request is meant to get everyone moved off of the old schedule structure to this new and improved one. This pull request is also meant to be a model for how the other containers will look like- how data is handled in a high level component, followed by a container component that manages ui state, followed by a series of functional components. 

I may be missing some key things like typescript, CSS-issues, features that I glossed over while merging the previous one with this one, so if you can help me with that, where I've missed that'd be great. Typescript and CSS-issues can be finished in another branch, but if I'm missing key features that was previously made that you feel is necessary here, lmk

src/components/Auth/AuthProvider.tsx
showduhtung: oh oh oh, do you mean the whole express redirecting the user to the react app?
Hmmm, so I had 2 reasons why I gave up on this. 
1) It's not that high of a security risk. I've talked to Ian about this- even if the user can "log themselves in", and even if they can forge themselves a token that bypasses them through `<PrivateRoute />`, they still need a valid token to retrieve any data to display any ui elements. Sure they can get html elements, but the data is protected. (Will need to confirm at a later date).

2) I think the solution to have Express reroute a React app only works if the React app and the Express app live on 2 different ports as completely separate apps. For @CedThen's sake (jk and all of ours), I have split this back up into two different apps, because the developer experience is majorly sucking. But when we are ready to push this to production, we will be merging the 2 apps BACK into 1 lol.. but this is for the sake of cost. I'm not 100% aware of the cost, but my guess is that hosting 2 apps cost 2x more than 1 app. Or at least around that number. Not all apps should be on one port- big businesses that have high scaling needs need their apps on separate ports, but because we don't have a fleet of apps, converging our webapp and serverapp into 1 is the most cost efficient. And I think this breaks the "redirect" method because of these lines in `server/index.ts`
```
const port = process.env.PORT || 8081;
const app: express.Application = express();
const DIST_DIR = path.join(__dirname, '../dist/');
const HTML_FILE = path.join(DIST_DIR, 'index.html');

app.use(express.static(DIST_DIR)); *
app.use('/api', require(...));

app.use('*', (req, res) => { **
   console.log('sending file');
   res.sendFile(path.join(__dirname, '..', 'dist/index.html'));
});
```

showduhtung: notice in the line where *, and ** are. Those blocks of code are commented out in our current app but in production, we want these code blocks uncommented out to tell our express serverapp to not only serve the web-server, but to statically serve the webpacked files of our ReactApp. 

Now when a `res.redirect('localhost:ReactAppPort8080')` is made, if our apps were separated, it has no problems redirect the app to that url. But when the apps are converged, we'd make the `redirect` to itself- `res.redirect('localhost:ReactAppAndServerPort8080')`, sending it to:
```
app.use('*', (req, res) => { 
   res.sendFile(path.join(__dirname, '..', 'dist/index.html'));
});
```
And then all it receives is a blank page. 
Honestly, I'm totally down if you can figure this out and see where I'm wrong. You'll probably need to play around with a different fullstack app that has both backend and frontend on the same port. But if you don't want, also ok

src/components/Auth/PrivateRoute.tsx
showduhtung: yee. 2nd time using it ever. but i think it's kinda magical

showduhtung: but I also think people can also find themselves in the rabbit hole of being too reliant on it everywhere. just like in redux

src/components/Schedule/NewScheduleForm.tsx
showduhtung: i love typescript. both informational and cleaner

src/components/Schedule/Schedule.tsx
showduhtung: Oo, what's the reasoning behind this? Ted and I were talking about how while developing, it's a tad bit annoying when there's much fluff nearer to the top.

src/components/Schedule/Schedule.tsx
showduhtung: Oh honestly this is just a temporary thing. I haven't really looked into how your showLoadingSpinner actually works. We can continue to compare the differences. I also noticed that the login screen wasn't changed away from CircularProgress that I implemented there, so we can also change it there as well.

src/components/Schedule/ScheduleTable.tsx
showduhtung: sure! will add

src/components/Schedule/utilities.ts
showduhtung: typescript caught tbh, i love typescript

src/query/apis/schedules.ts
showduhtung: wait you like the name "destroySchedule" or that I'm showing that we don't really need to destructure the data, just to restructure it later. Because destroySchedule sounds awesome

src/shared/styles/theme.js
showduhtung: material ui already has a `&:focus` styling with the underline underneath. to me, on the desktop it looks really confusing when the page opens and looks like this indicating that the first tab is selected:
Screen Shot 2020-12-30 at 4 16 51 PM

to this new kind of indicator that the 2nd tab is selected.
Screen Shot 2020-12-30 at 4 18 48 PM

showduhtung: Screen Shot 2020-12-30 at 4 20 02 PM
This picture to me feels consistent with the first picture

package.json
showduhtung: err I have no clue. But I don't think it's a breaking error?

src/components/Schedule/TableCell.tsx
showduhtung: oh I mean, yes some of it was ted, but also some of it was mine and yes WIP haha. ted actually wanted to go ahead and try his hand at rewriting and managing all the typescript stuff, but I'm gonna steal some of that from him

src/components/Auth/PrivateRoute.tsx
showduhtung: access- there will be more usages of the PrivateRoute to distinguish admin users and other users
exp- definitely needs to be implemented here in PrivateRoute

src/shared/styles/theme.js
hchiam: ah ok lgtm

src/components/Schedule/Schedule.tsx
hchiam: we can leave it as-is. i'd say it's fine either way, or depends. maybe i'm overapplying a rule, but it's like moving short guard clauses to the top of functions. if we agree on a style as a team, we can just go with that.

The thinking is that, instead of wondering what the else condition is, readers know immediately - this makes sense when the alternate condition is really short. So instead of this:

```js
if (condition) {
  // long
  // long
  // long
  // long
  // long
  // ...
} else { // huh? scroll back up to find the matching brace
  // short
}
```

and replace that with this:

```js
if (!condition) {
  // short
} else {
  // long
  // long
  // long
  // long
  // long
}
```

but thinking about it from the other point of view:

```js
if (condition) {
  // main
  // happy
  // path
  // goes
  // here
} else {
  // edge case you don't usually care about
}
```

showduhtung: hmm this makes sense to me. @CedThen what do you think? do you have a preference? If not, I'll change it, and we'll follow this perspective

src/query/apis/schedules.ts
alan-mj-lin: both

hchiam (approved):

alan-mj-lin (approved):

src/query/apis/services.ts
```diff
@@ -0,0 +1,8 @@
+import axios, { AxiosResponse } from 'axios';
+import { NewServiceData } from '../../shared/types';
+
+const secretIp = 'http://127.0.0.1:8081';
```
CedThen, line 4: not so secret is it now

PR #41, opened 2021-01-03
St/restructure

showduhtung:


showduhtung:
some things about this pull request. I'm starting to get to a point where I'm doing a lot of multiple api requests at a time, but am unable to because I can't pass an array as a query parameter:
https://stackoverflow.com/questions/49944387/how-to-correctly-use-axios-params-with-arrays

Fixing these endpoints means implementing these in: 
destroyUser
destroyEvent
updateScheduleAssignments

and fixing the endpoint route to handle multiple queries. 

Another thing I haven't done yet is migrate Teams to the container component like the others are. It's also currently not connected to the server and it's getting its data from database.ts. Anyone wanna take a crack at migrating this?

So the 2 tasks here are @alan-mj-lin @CedThen 
1) Fix endpoints to accommodate arrays as queries
2) Migrate Teams to use the container and add endpoints.

scripts/database/users.ts
```diff
@@ -1,14 +1,12 @@
 export const users = [
   {
-    id: 1,
```
showduhtung, line 3: I realized that this was the reason why autoincrementing wasn't working.

server/db/models/User.ts
```diff
@@ -11,21 +11,16 @@ const UserFactory = (
...
+    lastName: { type: DataTypes.STRING },
     email: {
       type: DataTypes.STRING(255),
       allowNull: false,
       unique: true,
     },
     password: {
       type: DataTypes.STRING(255),
-      allowNull: false,
+      allowNull: true,
```
showduhtung, line 23: when creating a user, passwords should be null. When we send an email and the user verifies their account, that will be where they create a new password.

src/components/Template/TemplateDisplay.tsx
```diff
@@ -51,6 +37,7 @@ export const TemplateDisplay = ({
           </TableRow>
         </TableHead>
         <TableBody>
+          {/* {data && console.log(data)} */}
```
CedThen: console log here

server/db/models/User.ts
hchiam: Maybe comment it if we're worried about forgetting this:
```suggestion
      allowNull: true, // create password upon verification
```

src/components/shared/ConfirmationDialog.tsx
```diff
@@ -10,14 +10,14 @@ import { buttonTheme } from '../../shared/styles/theme';
...
+  state: boolean;
   handleClick: (shouldDelete: boolean) => void;
 }
 
-export function ConfirmationDialog({ handleClick, isOpen, title }: SimpleDialogProps) {
+export function ConfirmationDialog({ handleClick, state, title }: SimpleDialogProps) {
   const classes = useStyles();
   return (
-    <Dialog onBackdropClick={() => handleClick(!isOpen)} open={isOpen}>
+    <Dialog onBackdropClick={() => handleClick(!state)} open={state}>
```
hchiam, line 20: why the rename from `isOpen` to `state` if we're keeping `open={state}`?

src/package.json
```diff
@@ -48,9 +43,7 @@
     "@types/material-ui": "^0.21.8",
     "@types/react": "^16.9.49",
     "@types/react-dom": "^16.9.9",
-    "@types/react-redux": "^7.1.9",
     "@types/react-router-dom": "^5.1.6",
-    "@types/redux-thunk": "^2.1.0",
     "@types/webpack": "^4.41.25",
     "awesome-typescript-loader": "^5.2.1",
     "css-loader": "^3.6.0",
```
hchiam, line 49: nice. less packages.

src/container/Members/Members.tsx
```diff
@@ -0,0 +1,69 @@
...
+    staleTime: 300000,
+    cacheTime: 3000000,
+  });
+
+  const createUser = useMutation<AxiosResponse<any>, AxiosError, NewUserData, unknown>(
+    addUser,
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries('users');
+        setIsSuccess('NewUser');
```
hchiam, line 33: `'NewUser'` would ideally be in an enum if we have time to refactor that here and in MembersContainer.tsx

src/container/Members/MembersContainer.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const targetRows: number[] = shiftKeyPressed
+      ? updateSelectedRows(lastSelected, id, data.users)
+      : [id];
+    if (selectedRows.includes(id))
+      // clause to unselect
+      setSelectedRows(selectedRows.filter((rowId) => !targetRows.includes(rowId)));
+    // clause to select
+    else setSelectedRows([...selectedRows, ...targetRows]);
+    setLastSelected(id);
+  };
```
hchiam, line 95: I personally find the innards of this `onSelect` function a little hard to understand. Maybe if we clean up this and [updateSelectedRows in utilities.ts](https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/41/files#diff-048f074a6238ac1846bff938834740ff799ade0b243947cb8f72f454e71da436R3), the bugs might be more obvious.

src/container/Members/utilities.ts
```diff
@@ -0,0 +1,23 @@
...
+  for (let i = 0; i < vectoredData.length; i++) {
+    if (startPushing) {
+      newRows.push(vectoredData[i].userId);
+      if (vectoredData[i].userId === end) break;
+    }
+    if (vectoredData[i].userId === start) startPushing = true;
+  }
+
+  // bug 1- hold the shift button down and continually select items. doesn't work as expected
+  // bug 2- this function only works on ordered id
```
hchiam, line 21: on my screen, the shift key ends up selecting text and only (de)selecting the one row I clicked on. Maybe we need to prevent an event from bubbling or happening. Or just need to refactor this utility function and [onSelect 
 in MembersContainer.tsx](https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/41/files/6506a10926fb884535e73be6ff4781497ad5877d#diff-01f664cf8394122c23e96d1fb3de83827754f9abba9c7ff15ae57b3088cacf66R85).

hchiam (changes requested):

src/container/Members/utilities.ts
hchiam: ~I'm trying it out locally to see if I can find a fix.~ -> https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/41#discussion_r551065615

src/container/Members/MembersContainer.tsx
showduhtung: mmmm this was actually a bugged feature that existed in the previous implementation. A fix for this bug isn't exactly in the scope of this pull request because we need to get everyone working on the same pages. What I'd suggest is that we move this to a ticket in jira- https://tjc-scheduling.atlassian.net/browse/TSA-134

hchiam: sounds good to me

showduhtung: Just a general idea of what this function is supposed to do, is that there is different expectations for when you press the shift key or not. When you don't press it, you'd expect individual rows to be selected or deselected. When you press it, you expect multiple rows to select/deselect.

src/components/Template/TemplateDisplay.tsx
showduhtung: console.log YOUR FACE

hchiam: https://memegenerator.net/img/instances/68620658/a-wild-consolelog-appeared.jpg

src/container/Members/Members.tsx
showduhtung: i'll make a ticket for someone to go through it all with a fine comb to enum everything

src/components/shared/ConfirmationDialog.tsx
showduhtung: I don't quite understand, `open={state}` is a change?

The reason for the change is more because if the name of a state is `isOpen`, then the name of the setState function that would go with it would be `setIsOpen`. And making a `setIsOpen(!isOpen)` is weirder than `setState(!state)` 

Then again it's kinda semantics.

PR #42, opened 2021-01-05
moved usemutations out, cleaned up some logic for error handling

CedThen:


src/routes/main.tsx
```diff
@@ -34,27 +34,27 @@ const Main = () => {
...
-                  <Template churchId={churchId} />
-                </Route>
-                <Route path="/members">
-                  <Members churchId={churchId} />
-                </Route>
-                <Route path="/">
-                  <Redirect to="/home" />
-                </Route>
-                {/* <Route>
+              {/* <QueryClientProvider client={queryClient}> */}
```
CedThen: is there a reason for two of these queryClientProviders?

src/components/Schedule/NewServiceForm.tsx
```diff
@@ -62,7 +63,7 @@ export const NewServiceForm = ({ onSubmit, onClose, error }: NewServiceFormProps
       <h2>Add A New Event</h2>
       {error && (
         <div style={{ color: 'red' }}>
-          {`Status code ${error.status}: ${error.message}`}
+          {`Status code ${error.response.status}: ${error.response.statusText}`}
```
hchiam, line 66: NewScheduleForm.tsx uses optional chaining, so I'd assume we should here too:
```suggestion
          {`Status code ${error?.response.status}: ${error?.response.statusText}`}
```

src/query/schedules.ts
```diff
@@ -1,4 +1,10 @@
 import { getScheduleAndData, getSchedules } from './apis/schedules';
+import { AxiosResponse, AxiosError } from 'axios';
+import { useMutation, useQueryClient } from 'react-query';
+import { DeleteScheduleData, NewScheduleData } from '../shared/types';
+import { postSchedule, destroySchedule } from './apis';
```
hchiam: currently unused here. for later?

src/container/Schedules/ScheduleContainer.tsx
```diff
@@ -148,9 +122,11 @@ export const ScheduleContainer = ({
       ? setSelectedEvents(selectedEvents.filter((id) => id !== eventId))
       : setSelectedEvents([...selectedEvents, eventId]);
 
+  console.log('rendering container', data);
```
hchiam: Remember to remove the [console log](https://memegenerator.net/img/instances/68620658/a-wild-consolelog-appeared.jpg) if it's not needed.

src/routes/main.tsx
hchiam: @showduhtung do you remember why? it got added [here](https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/39/files#diff-2f0ad119702516bef874bee8d4e51718094e5175db487993b88ec3404a557841R32) or [here](https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/39/commits/1d692ac4ae33a86294caa64439c94b236a912315) in main.tsx in PR #39

hchiam (changes requested):

src/container/Schedules/ScheduleContainer.tsx
CedThen: howard, you're too quick haha, i have some new pushes cleaning all of ur comments up xD

showduhtung:
@hchiam @CedThen , oh snap honestly I never noticed a double existence of ` {/* <QueryClientProvider client={queryClient}> */}`. Yeah, either positions work.

src/routes/main.tsx
CedThen: ok im deleting this one i commented out

src/container/Schedules/Schedule.tsx
```diff
@@ -10,49 +11,34 @@ interface ScheduleProps {
...
-      getScheduleData(
-        fetchedSchedules.length > 0 ? makeScheduleIdxs(tabs.data) : fetchedSchedules,
-      ),
+    () => getScheduleData(makeScheduleIdxs(tabs.data)),
     {
       enabled: !!tabs.data,
       refetchOnWindowFocus: false,
       staleTime: 100000000000000,
-      keepPreviousData: true,
+      keepPreviousData: true, // why is this true?
```
hchiam, line 29: I am not so strong in react-query :) @showduhtung ?

CedThen:
hm i'm not sure why but some of my later changes aren't showing up for me when im looking at files changed. might need to look at some individual commits to see them

src/container/Schedules/Schedule.tsx
```diff
@@ -10,49 +11,34 @@ interface ScheduleProps {
...
-  // is this dead?
-  // function fetchSchedule(tabIdx) {
-  //   if (tabIdx < tabs.data.length) {
-  //     setFetchedSchedule(tabIdx);
-  //     console.log(tabIdx, 'hello there');
-  //   }
-  // }
+  // only renders schedule once data is loaded. prevents excessive multiple re-rendering of the schedule itself
+  // maybe howard's css loading screen is better cause it handles being displayed/hidden better than mounting/unmounting react components?
+  if (!users.data) return <CircularProgress />;
```
hchiam, line 41: [We were only using `CircularProgress` as a temporary thing](https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/39#discussion_r550328475), so feel free to change it. CSS generally has better perf than DOM manipulation. And DOM manipulation triggers reflow, while simply toggling a class does not, especially with CSS `position: absolute`.

src/container/Schedules/ScheduleContainer.tsx
```diff
@@ -121,103 +123,91 @@ export const ScheduleContainer = ({ tabs, data }: ScheduleContainerProps) => {
       ? setSelectedEvents(selectedEvents.filter((id) => id !== eventId))
       : setSelectedEvents([...selectedEvents, eventId]);
 
+  // since the data check is handled in the parent component (where data is being queried), I think we should put the loading check there
   return (
     <>
-      {data?.schedules && data?.users ? (
```
CedThen: oh git thinks this is new, but all i did was move this line 126 up to schedules

src/components/Schedule/NewScheduleForm.tsx
```diff
@@ -9,12 +9,13 @@ import { buttonTheme } from '../../shared/styles/theme.js';
...
 import { stringLengthCheck } from '../../shared/utilities';
-import { HttpErrorProps, NewScheduleData } from '../../shared/types';
+import { NewScheduleData } from '../../shared/types';
+import { AxiosError } from 'axios';
 // TODO hook up teams with data from DB
 
 interface NewScheduleFormProps {
   onClose: (data: any) => void;
-  error?: HttpErrorProps;
+  error: AxiosError<any>;
```
showduhtung, line 18: Ah so actually I don't exactly agree with this. Because the error to be displayed will not only be an AxiosError, can be a validation error. We should keep the AxiosError and AxiosResponse handling in the apis, but for the UI to convert it into HttpErrorProps, although now that I'm looking at it, the current implementation is not obvious nor implied, and is poorly implemented by me.

src/components/Schedule/NewScheduleForm.tsx
```diff
@@ -151,18 +152,20 @@ export const NewScheduleForm = ({
...
-            </>
-          )}
+          {templates
+            ? templates.map(({ templateId: id, name }) => (
+                <MenuItem key={id} value={id}>
+                  {name}
+                </MenuItem>
+              ))
+            : [
+                <MenuItem value={1} key="1">
```
showduhtung, line 162: much better. But now that I'm looking at it, I don't think we should have any hardcoded default options. The default options should come from templates. 

Templates should be [...defaultTemplates, ...customizedTemplates]

I'll make a new ticket that addresses this

src/components/Schedule/NewServiceForm.tsx
showduhtung: Technically the previous method didn't need optional chaining as `error &&` handles it, but also in the previous method, I've made comments about using `error.response.{key}` at all

src/container/Schedules/Schedule.tsx
showduhtung: This is how react-query handles server-side pagination. While the ui still refetching for data, the server is not bothered. Our tabs is a form of pagination, especially over time when a church will have 10s of tabs.

src/container/Schedules/Schedule.tsx
showduhtung: fyi @hchiam, i pmed you on discord about loading logic to open a discussion about how we can handle it

src/components/Schedule/NewScheduleForm.tsx
showduhtung: @CedThen should i make a ticket to standardize how we handle errors? Because components that show errors should be able to handle both validation errors and data handlers in the same line. You can make the changes here, or we can move this to another ticket

showduhtung:
Hm.. looking at this pull request feels like it's counter to what I was trying to achieve which is to move all mutations out of ScheduleContainer, and that it would only have access to functions. Now every child component of scheduleContainer is tightly coupled to a particular api hook. Example is:
```
error={createService.error}
```

Error now has to be an AxiosError, when in fact may not be the case. We may want the error to be a validation error (my fault for not writing it up more clearly before). So if we wanted to solve this we'd have to abstract these functions higher up in the same component to be:

```
- const createSchedule = useCreateSchedule(setIsNewScheduleOpen);
+ const { createScheduleLoading, createScheduleError, createScheduleSuccess} = useCreateSchedule(setIsNewScheduleOpen);
```

This is essentially the exact same as it was before, just using the createSchedule.error elsewhere.

showduhtung (approved):

PR #43, opened 2021-01-06
Hc/component spinner

hchiam:
Added a new component-specific spinner `loadingTheme`. We can easily use it with `useEffect` and deps array `[]`, or with a `isLoading` ternary to toggle the class on/off. The spinner in `loadingTheme` automatically positions itself to the center of the element you add the styling to. In the gif below, I added an empty div when there's no schedules showing ([empty state](https://www.google.com/search?q=empty+state)).

Jan-06-2021 00-57-49

And commit https://github.com/House-on-a-Rock/tjc-scheduling-app/commit/c54328cd62d6148fd10d3107a8f1e18e68b8c2ec shows an example showing the placement of `loadingTheme` and `classes.loading` (you'll have to use a `loading` ternary).


showduhtung (approved):
looks all good to me. We may need more component specific loading css in the future. But we don't have the information for that now

hchiam:
Thanks @showduhtung - I'll just merge it now.

PR #47, opened 2021-01-12
Tc/manipulate events

CedThen:
In this PR, we can now add/remove events and services, and the autocompletes will display the correct options when a role assignment changes. 

Changes made are only local for now, I think in maybe 2 more PR it should be fully functional and can update the DB properly.


scripts/database/events.ts
```diff
@@ -5,126 +5,110 @@ export const events = [
     title: 'Usher',
     time: '10:15 AM',
     order: 1,
-    displayTime: true,
```
CedThen, line 8: This logic is handled client side now

hchiam:
@CedThen I fixed a minor merge conflict for you. For some reason GitHub couldn't figure out that I simply added a class onto `ScheduleContainer`. https://github.com/House-on-a-Rock/tjc-scheduling-app/pull/47/commits/e0b7c7af978eb9b34638565aab916b219b05ffbc

src/query/apis/services.ts
```diff
@@ -4,5 +4,6 @@ import { NewServiceData } from '../../shared/types';
 const secretIp = 'http://127.0.0.1:8081';
 
 export const postService = (serviceInfo: NewServiceData): Promise<AxiosResponse> => {
+  console.log('serviceInfo', serviceInfo);
```
hchiam: Remember to remove the [console log](https://memegenerator.net/img/instances/68620658/a-wild-consolelog-appeared.jpg) if it's not needed.

src/query/apis/teams.ts
```diff
@@ -0,0 +1,9 @@
+import axios, { AxiosResponse } from 'axios';
+import { secretIp } from '../../../secrets/secretStuff';
+// import { getLocalStorageItem } from '../../shared/utilities';
+
+export const getTeams = (churchId: number): Promise<AxiosResponse> => {
+  // const token = getLocalStorageItem('access_token')?.token;
+  // axios.defaults.headers.common.authorization = token;
+  return axios.get(`${secretIp}/api/roles?churchId=${churchId}`);
```
hchiam: remove comments or keeping for later just in case?

CedThen: howard you have the eyes of a hawk, spotting all this code i've forgotten about

src/components/Schedule/TimeCell.tsx
```diff
@@ -0,0 +1,37 @@
...
+  serviceIndex: number;
+}
+
+export const TimeCell = (props: TimeCellProps) => {
+  const { time, isDisplayed, onChange, rowIndex, serviceIndex } = props;
+  const [value, setValue] = useState<string>(time);
+
+  // React.useEffect(() => {
+  //   console.log('display', display);
+  // }, [display]);
```
hchiam: dead code?

src/container/Schedules/ScheduleContainer.tsx
```diff
@@ -135,68 +265,104 @@ export const ScheduleContainer = ({ tabs, data }: ScheduleContainerProps) => {
...
         />
         <Prompt
           when={isScheduleModified}
           message="You have unsaved changes, are you sure you want to leave? Unsaved changes will be lost"
         />
         {!data.schedules && <div style={{ height: '50vh' }}></div>}
         {/* {alert && <Alert alert={alert} unMountAlert={() => setAlert(null)} />} */}
-        {data.schedules?.map((schedule: ScheduleTableInterface, idx) => {
-          const { columns: headers, services: bodies, title, view } = schedule;
+        {dataModel.schedules.map((schedule: ScheduleTableInterface, scheduleIndex) => {
```
hchiam, line 291:
```suggestion
        {dataModel.schedules?.map((schedule: ScheduleTableInterface, scheduleIndex) => {
```
Otherwise I get a `dataModel.schedules is undefined` error.

src/container/Schedules/ScheduleContainer.tsx
```diff
@@ -63,67 +87,173 @@ export const ScheduleContainer = ({ tabs, data }: ScheduleContainerProps) => {
...
+      };
+    });
+    target.services = mutatedData;
+    setDataModel(dataClone);
+    retrieveChangesSeed(); // called just to update changesSeed.
+  }
+
+  function addService() {
+    const dataClone = { ...dataModel };
+    const target = dataClone.schedules[tab].services;
```
hchiam, line 174: I get that this is still a WIP, but just an FYI in case: I'm getting `Uncaught TypeError: dataClone.schedules is undefined` when the schedule tab has no services on it to begin with. I ran `npm run seed` in the root folder in your branch. Is it working for you when you try to add the first service to a schedule?

src/components/Schedule/useAutocompleteHook.tsx
```diff
@@ -0,0 +1,78 @@
+import { useState, useEffect, useRef } from 'react';
+
+export const useAutoCompleteHook = (
```
hchiam: Nice!  A custom hook.

src/container/Schedules/ScheduleContainer.tsx
```diff
@@ -63,67 +87,173 @@ export const ScheduleContainer = ({ tabs, data }: ScheduleContainerProps) => {
...
+      events: [],
+      serviceId: retrieveChangesSeed(),
+    };
+  };
+
+  function retrieveChangesSeed() {
+    // used to give temporary seeds to newly made (but unsaved) events/services. they are negative to distinguish from normal stuff from db
+    // also can keep be used to keep track of number of changes made, kinda
+    // its in a useRef right now, not sure if it needs to be there. But templateChanges will be used for a lot so idk
+    return templateChanges.current.changesSeed--;
```
hchiam, line 126: sounds hacky (using a negative number to track change count? while also making it the only negative ID in the DB to make it easier to find it in the DB?) but I assume this'll get cleaned up later since this PR is a first pass.

src/container/Schedules/ScheduleContainer.tsx
```diff
@@ -218,19 +384,20 @@ export const ScheduleContainer = ({ tabs, data }: ScheduleContainerProps) => {
...
+            //   scheduleId: tabs[tab].id,
+            //   order: dataModel.schedules[tab].services.length + 1, // need a better way to grab scheduleId and order
+            // })
+            addService()
           }
           onClose={() => setIsNewServiceOpen(false)}
           error={createService.error}
         />
-      </Dialog>
+      </Dialog> */}
```
hchiam, line 417: this is all commented out. dead code i presume?

src/container/Schedules/utilities.ts
```diff
@@ -119,3 +120,79 @@ export const memoizeData = (data: any) => {
   const makeDataLevel = () => data;
   return makeDataLevel();
 };
+
+// autocomplete cell functions
```
hchiam, line 119: this comment could get confusing if we accidentally add unrelated functions in between. maybe in one of the next PRs we can split/group these out into an `autocompleteUtilities.ts` and 
```js
import {
  extractRoleIds,
  getRoleOptionLabel,
  getUserOptionLabel,
  extractTeammateIds,
  teammates,
  blankTeammate,
  createBlankTask,
  createBlankEventCells,
  createBlankEvent,
  roleDisplay,
} from './autocompleteUtilities';
```

src/container/Schedules/utilities.ts
```diff
@@ -119,3 +120,79 @@ export const memoizeData = (data: any) => {
...
+    lastName: '',
+    email: '',
+    church: { name: '' },
+    churchId: churchId,
+    disabled: false,
+    teams: [],
+  };
+};
+
+const createBlankTask = (seedFx) => {
```
hchiam, line 162:
```suggestion
const createBlankTask = (seedFx: () => {}) => {
```
to avoid errors, since `seedFx` will only be used as a function, and with no parameters.

src/container/Schedules/utilities.ts
```diff
@@ -119,3 +120,79 @@ export const memoizeData = (data: any) => {
...
+  return {
+    taskId: seedFx(),
+    userId: seedFx(),
+  };
+};
+
+const createBlankEventCells = (cellLength: number, seedFx) => {
+  const taskCells: any = [{}, {}]; // data from these cells aren't actually being used, are just placeholders for rendering
+  for (let i = 2; i < cellLength; i++) {
+    // 2 (the columns for time/duty) is hardcoded here
```
hchiam: Explanatory variable to replace comment and magic number:
```suggestion
  const afterTimeAndDutyColumns = 2;
  for (let i = afterTimeAndDutyColumns; i < cellLength; i++) {
```

hchiam (changes requested):

src/container/Schedules/ScheduleContainer.tsx
CedThen: I think it might be used in the future, so i'm leaving it commented out instead of removing completely

src/container/Schedules/ScheduleContainer.tsx
CedThen: good catch, added some error handling to catch that

src/components/Schedule/DutyAutocomplete.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const [isCellModified, setIsCellModified] = useState<boolean>(false);
+  const [initialData] = useState<number>(dataId);
+
+  function onCellModify(isChanged: boolean, newValue: number) {
+    setIsCellModified(isChanged);
+    onChange(dataContext, newValue);
+  }
+
+  const tableCellClass = isCellModified ? classes.modified : classes.cell;
```
showduhtung: I feel like I see this more common inside className={}

General thought is no need to go and create constants unless they're used a lot. Takes storage.

src/components/Schedule/DutyAutocomplete.tsx
```diff
@@ -0,0 +1,146 @@
...
+  dataContext:        contains info like rowIndex/serviceIndex/roleId, used by onChange callback.
+  getOptionLabel:     string that shows in autocomplete
+  renderOption?:      basically just adds an icon to indicate which was the previously saved option, may be used to add more stuff
+  isSaved: boolean;   if initialData should update to the latest dataId
+*/
+
+interface AutocompleteCellProps {
+  dataId?: number;
+  dataSet?: any;
+  extractOptionId?: (data: any) => number[];
```
showduhtung: did this method have to be passed in via props? if it's just a pure function, it's probably a good idea to keep it as a utility service or something

showduhtung:
I noticed that in both autocomplete cells (TasksAutocomplete) and (DutyAutocomplete) you didn't handle inputValue. Was this intentional? I think iirc, we had a difference of opinion here if we should or not, and how we'd manage it because of cell focus and whatnot. If this was intentional, then yeah it's all good but I think it'd be cool to add

showduhtung:
This isn't in this PR but I found a bit of inconsistencies in your methods:

onTaskChange:
```
dataClone.schedules[tab].services[serviceIndex].events[rowIndex].cells[columnIndex].userId = newAssignee;
setDataModel(dataClone);
```

onAssignedRoleChange:
```
const targetEvent = dataClone.schedules[tab].services[serviceIndex].events[rowIndex];
targetEvent.roleId = newRoleId;
setDataModel(dataClone);
```

Does the onAssignedRoleChange work the same way as the first one? My guess is that targetEvent points a reference to dataClone.{...}, and then updates so should be okay?

src/components/Schedule/DutyAutocomplete.tsx
CedThen: made suggested change

src/components/Schedule/DutyAutocomplete.tsx
CedThen: moved these to imports instead of props

Issue #44, opened 2021-01-06
Howard todo's


hchiam:
backup notes to self, just in case, or if others need to know:

- [x] are you sure - confirm button red
- [x] "+" spacing left like chrome tabs
- [x] context menu styling
- [x] forms styling (#48)
  - [x] tooltips alignment
  - [x] input box spacing
  - [x] ~input grouping like in that article~ (not doing for the 3 forms - very simple and no clear grouping)
  - [x] confirm button should have colour
- [x] expandable (#49)
  - [x] columns aligning (might need to put invisible column match chevron size) [UPDATE: I went with custom code for now, since `<Collapse>` interferes with the level of table rows being direct children of the table]
  - [x] <https://material-ui.com/components/tables/#collapsible-table>

hchiam:
merged both PRs in just now before they get stale or before any new changes cause merge conflicts while i'm away

PR #48, opened 2021-01-18
Hc/style forms

hchiam:
"styling part 2": forms styling (#44)
- tooltips alignment
- input box spacing
- confirm button colour

# Before:
before

# After:
after

### DIFF
--- a/src/components/Schedule/NewScheduleForm.tsx
+++ b/src/components/Schedule/NewScheduleForm.tsx
@@ -173,7 +173,11 @@ export const NewScheduleForm = ({
         </div>
       </form>
       <div className={classes.buttonBottomBar}>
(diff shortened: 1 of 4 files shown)


src/components/Schedule/NewScheduleForm.tsx
```diff
@@ -133,40 +133,44 @@ export const NewScheduleForm = ({
             text="Select the begin date and end date for this schedule"
           />
         </div>
-        <ValidatedSelect
```
hchiam, line 136: it looks like a big diff, but all i did was wrap `<div className={classes.tooltipContainer}>` around each of the `<ValidatedSelect>`s that have a tooltip

hchiam:
merging this in now before it gets stale or before any new changes cause merge conflicts while i'm away

PR #49, opened 2021-01-25
Hc/expandable rows

hchiam:
### NOTE: I had to move `<TableBody>` out of ScheduleTable.tsx --> into ScheduleTableBody.tsx

collapsible-rows

[`<Collapse>`](https://material-ui.com/components/tables/#collapsible-table) doesn't work because it interferes with the "level" of table rows: rows need to be direct children of the table. It also turns out [you can group table rows by grouping them in separate `<tbody>`s, and mulitple `<tbody>`s can be inside a `<table>`](https://www.sitepoint.com/community/t/grouping-tr-table-rows/6966).

This is "styling part 3": expandable/collapsible rows (#44)


hchiam:
merging this in now before it gets stale or before any new changes cause merge conflicts while i'm away

PR #56, opened 2021-04-29
Tc/schedule refresh

CedThen:


server/routes/schedules.js
```diff
@@ -177,7 +173,7 @@ router.post('/schedule/update', certify, async (req, res, next) => {
         );
       }
     });
-    return res.status(200).send(`Schedule successfully updated`);
+    return res.status(200).send(`Schedule updated successfully!`);
```
showduhtung, line 176: lol

src/components/Schedule/ScheduleContainer.jsx
```diff
@@ -53,16 +68,28 @@ const ScheduleContainer = ({ churchId }) => {
...
+          {alert && (
+            <Alert alert={alert} isOpen={!!alert} handleClose={() => setAlert(null)} />
+          )}
         </div>
       )}
     </div>
   );
 
+  function onDeleteSchedule(tab) {
+    console.log(`tabClone, tab`, tabClone, tab);
```
showduhtung: console.log

src/components/Schedule/Table.jsx
```diff
@@ -97,56 +102,35 @@ const Table = ({
...
+                              </MuiCell>
+                              {cells.map((cell, columnIndex) => (
+                                <TableCell
+                                  cellIndices={{
+                                    serviceIndex: serviceIndex,
+                                    rowIndex: rowIndex,
+                                    columnIndex: columnIndex,
+                                  }}
+                                  roleId={roleId}
+                                  userId={cell.userId ?? 0}
```
showduhtung: I've learnt recently that it's better to keep conditional logic out of props as much as possible. You can try doing this in the TableCell component
```const TableCell = ({ ...props, userId = 0 }) => {```
This defaults `userId` to 0, if none is passed (but you'll have to check if `undefined` gets passed, I'm not sure if that affects this). 

But also only do this if this makes sense in other places we use TableCell (which I don't think we are).

File: src/components/Schedule/Table.jsx [hunk 4]
```diff
@@ -97,56 +102,35 @@ const Table = ({
...
+                                  taskId={cell.taskId ?? 0}
```
showduhtung: Same as above

src/components/Schedule/TableCell.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { renderOption } from './utilities';
+import { TimeCell, DutyAutocomplete, TasksAutocomplete } from '.';
+
+const TableCell = ({
+  cellIndices,
+  roleId,
+  userId,
```
showduhtung: so here you can make userId:
```userId = 0```

File: src/components/Schedule/useTasksAutocompleteHooks.js [hunk 6]
```diff
@@ -45,7 +43,23 @@ const useTasksAutocompleteHooks = (dataId, roleId, dataSet) => {
...
-  return [isCellModified, isCellWarning, managedDataSet, initialData];
+  useEffect(() => {
+    if (!isScheduleModified) {
+      // setIsCellWarning(false) this might be needed, have not tested it yet though
+      setIsCellModified(false);
+      setInitialData({ dataId, roleId, dataSet });
+    }
+  }, [isScheduleModified]);
+
+  const tableCellClass = isScheduleModified
```
showduhtung: ew nested ternaries. Try to use object literals?
https://betterprogramming.pub/dont-use-if-else-and-switch-in-javascript-use-object-literals-c54578566ba0

Even though that article says not to use if-else/switch, I think even if-else or switch is better here than nested ternaries.

showduhtung (approved):
Take a look at the comments I mentioned. Try them out, and if you're not successful, it's fine. I think this PR is fine either way. j

server/routes/schedules.js
CedThen: lol

src/components/Schedule/Table.jsx
CedThen: aw wanted to use nullish coalescing thingy but ok

PR #67, opened 2021-05-05
Tc/bug hunting

CedThen:


server/routes/schedules.js
```diff
@@ -25,6 +25,7 @@ router.get('/schedules', certify, async (req, res, next) => {
     const schedules = await db.Schedule.findAll({
       where: { churchId: churchId.toString() },
       attributes: ['id', 'title', 'view'],
+      order: [['id', 'ASC']],
```
showduhtung, line 28: is it not already default to ascending ids?

server/routes/schedules.js
```diff
@@ -124,8 +125,17 @@ router.post('/schedule', certify, async (req, res, next) => {
         });
       });
     }
+    // fetch tabs for return so it saves an api call
+    const schedules = await db.Schedule.findAll({
+      where: { churchId: churchId.toString() },
+      attributes: ['id', 'title', 'view'],
+      order: [['id', 'ASC']],
```
showduhtung, line 132: same as previous comment

src/components/Schedule/EditServiceForm.jsx
```diff
@@ -133,13 +133,15 @@ const EditServiceForm = ({ isOpen, onSubmit, onClose, serviceId, dataModel }) =>
       destination: { index: destination },
       source: { index: source },
     } = result;
-    if (!destination || destination === source) {
+    // idk why but using destination and source here messed it up
+    if (!result.destination || result.destination === result.source) {
```
showduhtung: oh you know what, gotta be careful about using numbers as truthy values. 
js interprets `!0` as `true`, and since we're using index, some index is probably `0`.

showduhtung: this is really weird that you say it works, because I feel like this statement will fail a lot. 
Screen Shot 2021-05-06 at 2 53 18 PM

showduhtung: I think you should take a closer look at what the result objects look like to make an accurate `if` statement

src/components/Schedule/ScheduleContainer.jsx
```diff
@@ -75,23 +79,35 @@ const ScheduleContainer = ({ churchId }) => {
...
 
-  function onDeleteSchedule(tab) {
-    // pbly need a more elegant solution in the future
+  function onCreateScheduleSuccess(res) {
+    setIsNewScheduleOpen(false);
+    setAlert(sendAlert(res));
+
+    const newTab = tabs.length;
+    setOpenedTabs((t) => [...t, newTab]);
+    setViewedTab(newTab);
```
showduhtung, line 90: maybe you can change viewedTab to currentTab. just a suggestion

src/components/Schedule/ScheduleContainer.jsx
```diff
@@ -75,23 +79,35 @@ const ScheduleContainer = ({ churchId }) => {
...
   );
 
-  function onDeleteSchedule(tab) {
-    // pbly need a more elegant solution in the future
+  function onCreateScheduleSuccess(res) {
+    setIsNewScheduleOpen(false);
+    setAlert(sendAlert(res));
+
+    const newTab = tabs.length;
+    setOpenedTabs((t) => [...t, newTab]);
```
showduhtung, line 89: for openedTabs, will we want the opened tabs to persist when we go throughout the app? maybe we wanna use useRef? (may not have any effect tbh since I think when the component unmount, the ref also resets)

src/components/Schedule/ScheduleContainer.jsx
```diff
@@ -75,23 +79,35 @@ const ScheduleContainer = ({ churchId }) => {
...
-    //   const next = i > 1 ? i - 1 : 0;
-    //   console.log(`next`, next);
-    //   clone.splice(i, 1, next);
-    //   console.log(`clone`, clone);
-    //   return clone;
-    // });
-    // setViewedTab((p) => (p > 0 ? p - 1 : 0));
+    setViewedTab(nextTab);
+    setOpenedTabs((t) => {
+      const clone = [...t];
```
showduhtung, line 99: this looks like it's always going to be organized in ascending order? like it looks like you're splicing the array to insert a tab inside. couldnt you just make a typical sort function that inserts an item in a list?

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -10,18 +10,15 @@ import { createStyles, makeStyles } from '@material-ui/core';
...
-// import { ContextMenu, ConfirmationDialog } from '../shared';
-
 import { processUpdate, createBlankService, formatData } from './utilities';
 import { updatedDiff } from 'deep-object-diff';
 
 import useScheduleMainData from '../../hooks/containerHooks/useScheduleMainData';
 
 // TODO
-// contextmenu functions don't work
+// contextmenu b r o k e n, but do we need it?
```
showduhtung, line 19: yes

showduhtung: but we don't have to have it for our first MVP.

src/components/Schedule/Table.jsx
```diff
@@ -107,27 +107,33 @@ const Table = ({
...
                             >
-                              <MuiCell
-                                align="left"
-                                style={{ display: 'flex', flexDirection: 'row' }}
-                              >
-                                <div {...provided.dragHandleProps}>
-                                  <ReorderIcon
+                              <MuiCell align="left">
+                                <div
+                                  style={{
```
showduhtung: can we remove inline stylings as much as possible?

showduhtung (changes requested):

server/routes/schedules.js
CedThen: it does not, i had cases where i got the data out of order, which is why i included this

src/components/Schedule/ScheduleContainer.jsx
CedThen: i think data is cached by react-query anyways. after testing it, there is no load screen when switching away then back to a schedule with already loaded tabs

PR #70, opened 2021-05-11
St/config dialogs

showduhtung:


src/components/Member/Dialogs/NewMemberFormDialog.jsx
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <CustomDialog
+      open={open}
+      title={title}
+      label="New-Member"
+      description={description}
+      handleClose={() => handleClick(CLOSE)}
+      handleSubmit={submitMember}
+    >
+      <FormField
```
CedThen, line 38: I think you opened an issue about converting these to validateTextFields?

src/components/Template/TemplateDisplay.jsx
```diff
@@ -34,8 +34,8 @@ const TemplateDisplay = ({ template, onAddClick }) => {
           </TableRow>
         </TableHead>
         <TableBody>
-          {data?.map((service) => (
-            <>
+          {data?.map((service, index) => (
+            <React.Fragment key={`${service.name}-${index}`}>
```
CedThen, line 38: don't we normally use <> </>? Why did you type out React.Fragment here

src/hooks/containerHooks/useMembersContainerData.js
```diff
@@ -13,15 +14,18 @@ const useMembersContainerData = (churchId) => {
   const createUser = useMutation(addUser, {
     onSuccess: () => {
       queryClient.invalidateQueries('users');
-      setIsSuccess('NewUser');
+      setState('CREATE_USER', 'SUCCESS');
```
CedThen, line 17: passed in 2 strings instead of an object?

PR #72, opened 2021-05-11
Tc/cell colors

CedThen:


server/utilities/helperFunctions.js
```diff
@@ -264,6 +264,12 @@ function areDatesEqual(d1, d2) {
   );
 }
 
+const cellStatus = {
+  SYNCED: 'synced',
```
showduhtung, line 268: keep the constants All Caps:

SYNCED= "SYNCED"

just the general way to showcase constants

server/utilities/helperFunctions.js
```diff
@@ -273,10 +279,16 @@ export async function populateServiceData(service, scheduleId, weekRange) {
   const eventData = await Promise.all(
     events.map(async (event) => {
       const { time, roleId, id: eventId, serviceId } = event;
+      const userRoles = await db.UserRole.findAll({
```
showduhtung, line 282: does this follow sequelize transactions that you mentioned before?

server/utilities/helperFunctions.js
```diff
@@ -298,16 +310,23 @@ export async function populateServiceData(service, scheduleId, weekRange) {
...
-async function retrieveTaskData(eventId, firstWeek, lastWeek) {
+async function retrieveTaskData(eventId, firstWeek, lastWeek, userIds) {
   const tasks = await db.Task.findAll({
     where: { eventId },
     attributes: ['id', 'userId', 'date'],
     order: [['date', 'ASC']],
   });
   const organizedTasks = tasks.map((task) => {
+    // checks if userId belongs to list of users with that role.
+    const doesRoleMatch = userIds.indexOf(task.userId);
```
showduhtung, line 321: this works without callback arguments?
`userIds.indexOf(id => id=== task.userId)`
?

server/utilities/helperFunctions.js
```diff
@@ -298,16 +310,23 @@ export async function populateServiceData(service, scheduleId, weekRange) {
...
     return {
       taskId: task.id,
       userId: task.userId,
+      // if role does not match, cell will receive 'warning' status and will appear red in frontend
+      status:
```
showduhtung, line 326: hmm so this isn't being stored in the db? wouldn't it be easier functionally if we stored the state? why is this better?

src/components/Schedule/Table.jsx
```diff
@@ -211,20 +213,19 @@ const Table = ({
...
-
     dataClone[serviceIndex].events.splice(rowIndex, 1);
-
     setDataModel(dataClone);
-    retrieveChangesSeed(); // called just to update changesSeed.
   }
 
   // onChange Handlers
-  function onTaskChange(dataContext, newAssignee) {
+  function onTaskChange(dataContext, newAssignee, initialId) {
```
showduhtung, line 222: learnt this one from hain-lee, for future use, make the argument list length of 1, and destructure:

```
function onTaskChange({dataContext, newAssignee, initialId}) {
```

more work, but it's self-documenting

src/components/Schedule/TasksAutocomplete.jsx
```diff
@@ -30,26 +28,23 @@ const TasksAutocomplete = React.memo((props) => {
...
   const classes = useStyles();
+  const [initialId, setInitialId] = React.useState(dataId);
 
-  const [tableCellClass, managedDataSet, initialData] = useTasksAutocompleteHooks(
-    dataId,
-    roleId,
-    options,
-    isScheduleModified,
-  );
+  React.useEffect(() => {
```
showduhtung, line 40: ah I think this was what I was talking about where the hook reads the changes

src/components/Schedule/TasksAutocomplete.jsx
```diff
@@ -59,16 +54,13 @@ const TasksAutocomplete = React.memo((props) => {
...
         renderOption={(option) =>
-          renderOption(
-            getUserOptionLabel(option, managedDataSet),
-            option === initialData.dataId,
-          )
+          renderOption(getUserOptionLabel(option, options), option === initialId)
         }
         value={dataId}
-        onChange={(event, newValue) => onChange(dataContext, newValue)}
+        onChange={(event, newValue) => onChange(dataContext, newValue, initialId)}
```
showduhtung, line 63: been reading that it's a bad idea to pass nameless functions into props directly. mainly because it's not self-documenting.

you might notice my config dialogs, I try to name them before passing in as props. except my dialogconfig which the key itself is the name. 

except direct setStates, it's pretty pointless to create a named function just to setState imo

showduhtung: not an impt comment

src/components/Schedule/TasksAutocomplete.jsx
showduhtung: not an impt comment

src/components/Schedule/utilities.jsx
```diff
@@ -167,14 +167,16 @@ export function formatData(dataModel, previousServices) {
   return updated;
 }
 
+// Styling for dropdown box for task autocompletes
```
showduhtung, line 170: is this a todo comment?

if so write, // TODO

vscode actually picks this up and changes the color of the comments

showduhtung (commented):
just talk to me about why status isnt being saved in tasks. i think it's the right way to go

server/utilities/helperFunctions.js
CedThen: my thinking is that it only really affects the visual rendering of the table right now, and i can make do without the extra column. if we find we need it later for functionality reasons, we can always add it later imo

showduhtung: Ah actually tHIS was the comment. Yah I think we'll definitely be coming around to this some time. Keeping the state in db would allow for true single source of truth that doesn't getting messed up and whatnot. But yah we can do that later

PR #73, opened 2021-05-12
Tc/error checking

CedThen:


server/utilities/helperFunctions.js
```diff
@@ -393,7 +393,7 @@ export const updateEvents = async (events, t) => {
           parentSchedule.end,
           parentService.day,
         );
-        taskDays.forEach((date) =>
+        taskDays.forEach(async (date) =>
```
showduhtung: dont think this needs to be async, a callback function is inherently either asynchronous or synchronous
https://dev.to/marek/are-callbacks-always-asynchronous-bah

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -32,6 +28,9 @@ const ScheduleMain = ({
 }) => {
   const classes = useStyles();
   const [isScheduleModified, setIsScheduleModified] = useState(false);
+  const [isScheduleWarning, setIsScheduleWarning] = useState(false);
```
showduhtung, line 31: doesn't look like this is being used anywhere, but if you do use this (which you probably will) you can use the CustomDialog once, and compose the warning state and the dialog state together. 

using `const [dialogState, setDialogState] = useState('')`

where the dialogState can be `SCHEDULE_WARNING`, or `IMPROPERLY_ASSIGNED_CELLS`

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -49,22 +48,27 @@ const ScheduleMain = ({
     if (schedule) setDataModel(ld.cloneDeep(schedule.services));
   }, [schedule]);
 
-  const outerRef = useRef(null);
+  // used to track isScheduleModified
+  useEffect(() => {
+    if (templateChanges.current.changesSeed < -1) setIsScheduleModified(true);
```
showduhtung: couldve probably done 
`setIsScheduleModified(templateChanges.current.changesSeed < -1)`

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -76,43 +80,70 @@ const ScheduleMain = ({
...
+      ></CustomDialog>
     </div>
   );
 
-  function onSaveScheduleChanges() {
+  function isWarningStatus(data) {
+    let isWarning = false;
+    data.forEach((service) => {
+      service.events.forEach((event) => {
+        event.cells.forEach((cell) => {
```
showduhtung: so basically this is if there are any warnings at all, warning should be turned on

`event.cells.find()` probably works more optimally here, because it finds the first value and then stops. compared to `forEach` which does every single iteration

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -76,43 +80,70 @@ const ScheduleMain = ({
...
+    // check if any cells have status: warning
+    const isWarning = isWarningStatus(dataModel);
+    setIsDialogOpen(isWarning);
+
     const diff = updatedDiff(schedule.services, dataModel);
-    // need error checking before running diff... or do we
     const processedDiff = processUpdate(diff, dataModel);
     updateSchedule.mutate({ tasks: processedDiff });
-    // setIsScheduleModified(false);  make this an onsuccess?
+    resetChangesSeed();
```
showduhtung, line 115: if you're doing a `onSaveSchedule`, you should stop the function at `isWarning` and not go to `resetChangesSeed`
 or is there a reason why `resetChangesSeed` needs to be reset?

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -76,43 +80,70 @@ const ScheduleMain = ({
         churchId={churchId}
         isScheduleModified={isScheduleModified}
         setIsScheduleModified={setIsScheduleModified}
-        retrieveChangesSeed={retrieveChangesSeed}
+        incrementChangesSeed={incrementChangesSeed}
```
showduhtung, line 82: can you explain to me changesSeed? what is seed?

src/components/Schedule/TableCell.jsx
```diff
@@ -8,7 +8,7 @@ const TableCell = ({
   roleId,
   userId = 0,
   taskId = 0,
-  status = cellStatus.SYNCED,
+  status,
```
showduhtung: why was this removed?

showduhtung (commented):
just a couple questions on what's going on, I've noted them in the comments below.

server/utilities/helperFunctions.js
CedThen: kk, i'll make this change on  my reorg branch cause i moved this function to a different file

src/components/Schedule/ScheduleMain.jsx
CedThen: kk

src/components/Schedule/ScheduleMain.jsx
CedThen: yeah this part was unfinished, i shudda put wip or something
felt like PR was getting too big at this point

src/components/Schedule/ScheduleMain.jsx
CedThen: its an integer i use to keep track of the number of changes being made. kept in a useRef to prevent it from causing re-renders. it needs to be redone a little bit i think

src/components/Schedule/TableCell.jsx
CedThen: wasn't necessary with my solution but i can add it back

src/components/Schedule/ScheduleMain.jsx
CedThen: anddd im pbly gonna get rid of it, it kinda sucks

PR #76, opened 2021-05-18
Tc/update2

CedThen:
Switching days when editing the service now updates all the tasks associated with that service with the appropriate date. If the number of tasks/event/service changes, it will now automatically delete/create tasks to fit the number of actionable days in the schedule period

### DIFF
--- a/server/services/schedules.js
+++ b/server/services/schedules.js
@@ -197,6 +197,7 @@ async function updateModel(model, t) {
             );
           } else {
             // create new events
+            console.log('creating new event');
             const newEvent = await db.Event.create(
               {
                 time,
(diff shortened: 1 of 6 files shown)


CedThen:
this one first

server/services/schedules.js
```diff
@@ -146,64 +146,142 @@ export const updateSchedule = async (changes) => {
...
-        return db.Service.create(
-          { name: name, day: day, scheduleId: scheduleId, order: index },
-          { transaction: t },
-        );
+    events.map(async (event) => {
+      const tasks = await db.Task.findAll({
+        where: { eventId: event.id },
+        order: [['date', 'ASC']],
+      });
+      switch (tasks.length - taskDays.length) {
```
showduhtung, line 237: switch case may not be super necessary when there are for sure 2 results

server/services/schedules.js
```diff
@@ -146,64 +146,142 @@ export const updateSchedule = async (changes) => {
...
+        case -1:
+          if (isSameWeek(tasks[0].date, taskDays[0])) {
+            await db.Task.create({
+              date: taskDays[taskDays.length - 1],
+              eventId: event.id,
+            });
+            await Promise.all(
+              tasks.map(async (task, index) =>
+                task.update({ date: taskDays[index] }, { transaction: t }),
```
showduhtung, line 246: oh what is transaction?

showduhtung: or how is it being used?

src/components/Schedule/utilities.jsx
```diff
@@ -49,18 +49,18 @@ export const blankTeammate = (churchId) => {
...
     disabled: false,
     teams: [],
   };
 };
 
-export const createBlankEvent = (serviceId) => {
+export const createBlankEvent = (serviceId, incrementChangesCounter) => {
   return {
-    eventId: null,
+    eventId: -incrementChangesCounter(),
```
showduhtung, line 61: maybe instead of `-function()`, maybe `function counter(type = "down" || "up")`

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -34,6 +34,7 @@ const ScheduleMain = ({
   const [isScheduleModified, setIsScheduleModified] = useState(false);
   const [dialogState, setDialogState] = useState({ isOpen: false, state: '' });
   const [isEditMode, setIsEditMode] = useState(false);
+  const changesCounter = useRef(0);
```
showduhtung, line 37: ohh i get what's going on now. maybe you can make this a hook. i can see this easily shareable

that includes resetChangesCounter and incrementChangesCounter

src/components/Schedule/ScheduleMain.jsx
```diff
@@ -129,6 +130,7 @@ const ScheduleMain = ({
         churchId={churchId}
         isScheduleModified={isScheduleModified}
         setIsScheduleModified={setIsScheduleModified}
+        incrementChangesCounter={incrementChangesCounter}
```
showduhtung, line 133: my guess is that this determines the toolbar states?

you could maybe call it changesState

showduhtung (approved):
just the one comment and suggestion to abstract into a hook. otherwise lgtm

PR #77, opened 2021-05-18
tc/data-access

CedThen:
added data-access file where i extracted out all the db.Whatever.action calls so they're more easily reused

### DIFF
--- a/server/routes/schedules.js
+++ b/server/routes/schedules.js
@@ -1,11 +1,10 @@
 /* eslint-disable consistent-return */
 import express from 'express';
 
+import { findAllChurchSchedules, destroySchedule } from '../services/dataAccess';
 import {
   createSchedule,
-  deleteSchedule,
   doesScheduleExist,
-  retrieveChurchSchedules,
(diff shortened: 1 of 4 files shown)


CedThen:
other one first

server/routes/schedules.js
```diff
@@ -16,7 +15,7 @@ const router = express.Router();
 router.get('/schedules', certify, async (req, res, next) => {
   try {
     const { churchId } = req.query;
-    const schedules = await retrieveChurchSchedules(churchId);
+    const schedules = await findAllChurchSchedules(churchId);
```
showduhtung, line 18: I'm assuming these are just naming changes? lgtm

server/services/dataAccess.js
```diff
@@ -0,0 +1,96 @@
+import db from '../index';
```
showduhtung, line 1: like a lot

showduhtung:
looks like an abstraction pr right? looks real good to  e

